Issue #29, opened 2026-07-18
docs: work package sets a >=2 reviewer bar where ASSURANCE.md Tier-3 requires a hard floor of 3


zoolok17:
## Summary

`docs/WORK-PACKAGE-native-work-spine.md` sets a review bar **below** the one `docs/ASSURANCE.md` mandates for the tier this feature actually falls into. Two teams following their respective documents will disagree about whether the work is releasable, and the discrepancy surfaces at merge rather than at planning.

## The contradiction

`docs/WORK-PACKAGE-native-work-spine.md:113`:

> **Adversarial review before merge:** ≥2 cross-family reviewers on the final SHA + your lead gate + CI.

`docs/ASSURANCE.md:72-78` defines **Tier 3 (adversarial cross-family DESIGN panel)** as covering a diff that touches *any* of: authority/permission/approval/gate; provenance/integrity; fail-open/fail-closed error semantics; durability/atomicity/recovery; persistent-state contracts; external side-effect boundaries; or normative policy.

`docs/ASSURANCE.md:93-96` then sets, for Tier 3, **a hard floor of 3 eligible independent reviewers** (≥2 model families; distinct predeclared lenses; no designer/builder/lead counted; ephemeral evidence-only reviews may add counters but do **not** satisfy the minimum).

The Native Work & Evidence Spine matches Tier 3 on at least five of those dimensions simultaneously — it *is* a gate/authority surface, it *is* a provenance and persistent-state contract, and its whole design problem is fail-open semantics. So the governing floor is 3, not 2, and the work package understates it.

`ASSURANCE.md:63` ("Uncertainty rounds UP") points the same way.

## Related, smaller

`docs/WORK-PACKAGE-native-work-spine.md:104` (§6) instructs adding one subparser line to `cli.py`. Per `ASSURANCE.md:79-82` a public contract change (CLI flag, JSON schema, config key) is **Tier 2 minimum** on its own — and the feature's entire `.agenttalk/work/` + `.agenttalk/artifacts/` JSON surface is a public contract too. Worth stating explicitly in the work package so the tier is not re-litigated per deliverable.

## Suggested fix

Amend the work package §7 to name Tier 3 and the 3-reviewer floor directly, and note that the RFC (D1) itself is the Tier-3 *design* panel — i.e. the floor applies to the design review, not only to the final code diff.

## What we did on the second-laptop team

[...]

Issue #28, opened 2026-07-18
supervise --select-pwsh: bare form re-probes instead of returning the recorded host, and no non-admin install path is documented


zoolok17:
## Summary

Two related problems make the supervisor unreachable for a non-admin Windows operator, and then make the documented "get the host path" idiom fail even after the host is correctly recorded.

## 1. `supervise --select-pwsh` with no argument re-probes instead of returning the recorded selection

`docs/supervisor-tutorial.md:298-302` documents this as the canonical way to obtain the host:

```powershell
$pwshPath = (agenttalk supervise --select-pwsh | ConvertFrom-Json).path
& $pwshPath -NoLogo -NoProfile -NonInteractive -File .\.agenttalk\supervisor.ps1
```

On a portable/nonstandard install this **fails**, because the bare form re-runs auto-discovery (which considers only canonical `C:\Program Files` locations) rather than returning the already-recorded, already-validated selection:

```
agenttalk supervise --select-pwsh: PowerShell Core host selection failed:
cannot open candidate (winerror 3)
```

…even though the selection is valid and `doctor` simultaneously reports:

```
[ok  ] powershell_host   C:\Tools\PowerShell\7\pwsh.exe (explicit, Core 7.6.3, age=0s)
```

So the documented launch snippet cannot be copy-pasted by anyone whose host is not in Program Files. The working substitute is to read the record directly:

```powershell
$pwsh = (Get-Content .\.agenttalk\powershell-host.json -Raw | ConvertFrom-Json).path
```

**Suggested fix:** when a valid recorded selection exists and no `--pwsh` is given, return it (re-validating identity) instead of re-probing. Failing that, document the `powershell-host.json` read as the supported idiom for portable installs.

## 2. No documented path for a non-admin operator

`winget install Microsoft.PowerShell` on Windows 11 yields the **MSIX/Store** build under `C:\Program Files\WindowsApps\...`, which agenttalk explicitly refuses:

```
PowerShell Core host selection failed: WindowsApps aliases are unsupported;
select the real local pwsh.exe
```

The refusal is correct. But the remediation text — "install PowerShell Core 7+ from https://aka.ms/powershell" — points at an MSI that needs elevation, and auto-discovery only looks in Program Files. A non-admin operator following the docs is stuck with no stated way forward, even though a fully supported one exists.

**What worked** (no admin, no registry, no PATH change, reversible by deleting one folder):

```powershell
# official portable zip
[...]

Issue #27, opened 2026-07-18
supervisor: never-launched agent is classified STUCK_OR_DEAD (kill+resume) on first poll instead of a clean launch


zoolok17:
## Summary

On its **first poll**, the supervisor classifies a configured-but-never-launched agent as `STUCK_OR_DEAD` and plans `stuck_recover: heartbeat stale; best-effort kill + resume`, rather than a clean initial launch. `launch_grace_seconds` does not protect this path because grace is only evaluated relative to a *recorded* launch, and on cold start there is none.

It converges — the agents come up healthy seconds later — so this is a correctness-of-classification and observability bug, not an outage. But it misreports the single most common operation (starting a team) as a stuck-agent recovery.

## Reproduction (two ways, both observed)

**A. Cold start of a fresh team.** Supervisor process start `2026-07-18T14:16:18.384Z`. First entries in `.agenttalk/state/supervisor-events.jsonl`:

```
2026-07-18T14:16:19Z | claude-dev | stuck_recover | STUCK_OR_DEAD
2026-07-18T14:16:19Z | claude-rev | stuck_recover | STUCK_OR_DEAD
2026-07-18T14:16:19Z | codex-dev  | stuck_recover | STUCK_OR_DEAD
2026-07-18T14:16:19Z | codex-rev  | stuck_recover | STUCK_OR_DEAD
2026-07-18T14:16:41Z | claude-dev | none          | HEALTHY_IDLE
...
```

That is ~0.6s after process start, against four agents that had never been launched. All four were healthy 22s later — far inside the configured `launch_grace_seconds: 120`, so grace was never the limiting factor; it simply did not apply.

**B. Minimal, on-demand.** With the supervisor already running healthy (`poll 40 - 4/4 agents healthy`), add one new agent block to `supervisor.json`:

```
supervisor: codex-sec: STUCK_OR_DEAD -> stuck_recover: heartbeat stale; best-effort kill + resume
supervisor: codex-sec: HEALTHY_IDLE (healthy)
```

One agent, never launched, immediately "stale + kill".

## Analysis

[...]

Issue #26, opened 2026-07-18
docs: wrapped-Codex stuck_after_seconds guidance is below the watchdog-preempt floor, silently disabling stale recovery


zoolok17:
## Summary

`docs/supervisor-tutorial.md` and `README.md` recommend a wrapped-Codex `stuck_after_seconds` **below the floor at which the planner silently revokes restart authority**. An operator who follows the documented guidance ends up with a wrapped Codex agent that is **never auto-recovered when it wedges** — with no error, and no indication in `doctor` or `bootstrap-check`.

Found while standing up a supervised Claude+Codex team on Windows against `master` (source 0.78.1).

## The disagreement

| Source | Says |
|---|---|
| `docs/supervisor-tutorial.md:211` | example config uses `"stuck_after_seconds": 1200` |
| `docs/supervisor-tutorial.md:251-254` | "wrapped **Codex** ... default 900s, and for a heavy reasoning/review role pick **1200–1800s**" |
| `docs/supervisor-tutorial.md:527` / `README.md:1287` | wrapped-Codex default is 900s |
| `src/agenttalk/supervisor.py:377` | actual default is **2400** |
| `src/agenttalk/supervisor.py:3749-3772` | restart-on-stale **degrades to warn-only** when `stuck_after_seconds <= turn_watchdog.turn_elapsed_seconds + 300` |

With the shipped `turn_watchdog.turn_elapsed_seconds` default of 1800, the effective floor is **2101s**. Every value the tutorial recommends (1200, 1800) is below it, as is the default it claims (900).

The guardrail note at `supervisor-tutorial.md:255-256` only documents the older `< 600s` floor and predates the watchdog-preempt guard added with the per-turn watchdog.

## Why it matters

The value is deliberately never silently coerced — which is right — but the *consequence* is silent from the operator's side: the agent still launches, still heartbeats, still shows `HEALTHY_IDLE`, and `supervise --bootstrap-check` reports no error. The loss of restart authority only shows up as an agent that stays wedged forever.

This is the failure mode the threshold exists to prevent, reintroduced by following the docs.

## Suggested fix

1. Correct `supervisor-tutorial.md:211`, `:251-254`, `:527` and `README.md:1287` to the real default (2400) and state the derived floor as `turn_elapsed_seconds + 300`.
2. Replace the stale `< 600s` guardrail paragraph with the watchdog-preempt rule.
[...]

PR #30, opened 2026-07-18
D1: Native Work & Evidence Spine RFC (draft — for cross-team read)

zoolok17:
## D1 — Native Work & Evidence Spine RFC

Design-first deliverable **D1** from `docs/WORK-PACKAGE-native-work-spine.md`, built by the independent second-laptop team. **Draft PR — for the cross-team read, not to merge.**

`agenttalk work` binds existing primitives (domains, lanes, gates, closes, review threads, onboarding) into one durable delivery record. It is a **link-and-project layer**: it never stores a second copy of a truth another module owns, status and verdict are projections that are never persisted, and `work.compute_verdict` is pure over CLI-resolved bundles.

### Status: ratified, closed, and being built against

| | |
|---|---|
| **Ratified at** | `fafd8fd` / blob `9515d6ae` · 3369 lines |
| **Reviewers** | `claude-rev` (boundaries/schema) · `codex-rev` (fail-open) · `codex-sec` (security/authority/provenance) |
| **Architect** | cross-read and approved — *"exceeds the work-package bar… cleared to build D2"* |
| **Rounds** | 5 review rounds → chain closed by unanimous vote → **13 amendments** |
| **Convergence (D1)** | 11 consecutive rounds with zero new defect classes — **this streak has since broken in D2; see below** |

Three independent lenses across two model families, distinct predeclared lenses, author and lead excluded — meeting the `ASSURANCE.md` Tier-3 floor of 3 (see #29, where the work package's "≥2" was corrected at source).

### The result worth reading this PR for

D1 reached zero P0s through five adversarial review rounds. The three lenses then **voted unanimously to close the review chain**, on the argument that remaining risk was better found by building than by reading.

Thirteen amendments followed. **Essentially every defect in them came from implementing or testing the spec, not from re-reading it** — because a gap makes no false claim, so there is no wrong sentence for a reviewer to catch.

Examples of what only building found:
- The RFC **mandated `store._config_lock()`** — a call our own ownership boundary forbids. Three lenses and the architect all approved that document.
- **Circular validation**: `work deliver` would have passed an artifact's own `delivered_head` into its validator — asking whether the artifact agrees with itself, which every well-formed artifact answers yes to, *including a stale one*.
- **Omission-as-silent-waiver at four levels** — item, classification, child collection (`checks: []`), and finally the predicate domain itself (a zero-path diff makes an "outside scope" predicate vacuous).
- Two **dead enum members** with no producer in any phase, one of which made a state-table disjunction read as having two satisfiable branches when it had one.

### Honest limits — claims withdrawn rather than overstated

- **`work_waiver_backed_close` deleted.** `close.py`'s published record stores blocker *names* only, so work cannot detect a waiver-backed close. All four invisible routes documented; scheduled to D6 blocking D4.
- **`local_operator` and `automation_ci` unreachable in D1–D5**, with the consequence stated plainly: under the default floor no release-blocking item can reach GO.
- **`min_independent_roots > 1` unsatisfiable**, because claimed execution roots are self-asserted without a launch-time nonce.
- **Safety Invariant 22's replay claim narrowed** to what the schema can actually implement.

### ⚠ Cross-team dependency

**D6 item 7 blocks D4 on a close-provenance envelope requiring `close.py` changes** — outside this team's ownership boundary, filed as a request per work package §2. The architect has confirmed it is owned, specced to our four invisible routes *and* the independent-source rule, and lands before D4.

### Method that outlived the document

Five **domain-blind mechanical sweeps**, each added because a defect escaped the previous set — consistency, withdrawal, reachability, consumer-propagation, circular-validation. Plus: *any normative surface indexed on a non-topical axis goes stale as a property of the index, not of diligence* (four such surfaces stale at least once; topical sections zero times across thirteen amendments).

The architect has adopted the circular-validation rule for `close.py` and the lead-listening finding for their agent-lifecycle RFC.

### Scope

Four documentation files. **No `src/`, no `cli.py`, no `store.py`, no `wrapper/`** — the ownership boundary held for the whole deliverable. The single `cli.py` dispatch line is filed as a request and lands with D2.

`docs/TEAM-SETUP-native-work-spine.md` reproduces the supervised Claude+Codex team from a bare clone. `docs/RFC-native-work-spine-ARCHITECT-RULINGS.md` carries the architect's verbatim rulings so the RFC's citations are checkable in-repo.

### Related upstream issues

#26 (wrapped-Codex `stuck_after_seconds` below the watchdog floor) · #27 (never-launched agent classified `STUCK_OR_DEAD`) · #28 (`--select-pwsh` re-probes; no non-admin path) · **#29 (reviewer floor — fixed at source)** · #31 (divergent canonical-hash conventions) · #33 (deadman configured-but-not-running is silent)

---

## Update — the convergence streak broke in D2 (2026-07-19)

The D1 figure above is a true historical statement about the D1 chain. It is **no longer the current status**, and leaving it unqualified would misrepresent the work, so:

**Two new defect classes have since been found, both during D2 review.**

**`descriptor-validated-as-instance`** — running the checker built for a *member* of a set against a *descriptor* of that set. Domain containment was proved by passing scope **globs** to `domains.check_paths`, which classifies a *concrete path*. A domain owning `src/*` accepts scope `src/**` (because `*` matches the literal segment `**`) while that scope's own member `src/deep/file.py` lies outside the domain. Found independently by two reviewers in different model families, reproduced a third time before disposition.

**`valid-input algorithmic exhaustion`** — a specified predicate that is exponential on shape-valid input. The replacement containment rule's `**` recurrence has overlapping subproblems and is mandated on *every read* with no memoization or bound. Measured call counts: 335 / 4,718 / 68,067 / 999,361 / 14,857,999 at n=4/6/8/10/12, and 223 million at n=14. Two independent transcriptions of the spec produced identical counts, which is what establishes that the *specification* is exponential rather than one implementation being wrong.

**Why this update exists at all.** A convergence metric that only ever improves is not measuring convergence. The streak was reported here when it was running; reporting the break is the same obligation. Both classes were found by the review structure working as intended — in each case by the *third* reviewer, whose lens the other two did not cover.

**Structural datum worth recording:** on the D2 increment-2 review, the two Codex lenses each found a release-blocking P0 that the other missed, and both independently found the same new-class defect. Had the panel been run at the work package's original minimum of two reviewers, *which two were chosen* would have decided which P0 shipped. That is the most concrete argument for the `ASSURANCE.md` Tier-3 floor of three that this project has produced.

The D1 RFC at `fafd8fd` / blob `9515d6ae` is unaffected — both classes were found against D2 implementation and against a later amendment, not against the ratified D1 text.

### DIFF
--- a/docs/REVIEW-SWEEPS.md
+++ b/docs/REVIEW-SWEEPS.md
@@ -212,16 +212,40 @@ This is the only sweep of the six that assumes the answer to "did we check?" is
 YES and asks a further question. Every checklist is phrased as did-we-check;
 every instance below passed a real check, honestly performed.
 
-### Two species, with different defences — both must be run
+### Three species, with different defences — all must be run
 
 - **WRONG IDENTITY.** The check touched a different object than the claim.
   *Defence:* name the object, and label its TYPE.
-- **WRONG TIME.** The check touched the right object, but the object moved
-  between the check and the action it authorised.
+- **WRONG TIME — the object moved.** The check touched the right object, but
+  it changed between the check and the action it authorised.
   *Defence:* re-check AFTER the action, or remove the concurrency.
   > You can aim correctly and still miss, if the thing moves.
+- **WRONG TIME — the WORLD moved.** The object did not change at all. Its
+  *description* went stale, because everything around it changed.
+  *Defence:* re-derive a preserved artifact against the CURRENT world before
+  use. **Its name is not evidence.**
 
-A sweep naming only the first species passes the second cleanly. The general
+The third is the most dangerous of the three, and the reason is the timescale.
+The first two have a window measured in minutes. This one has a window measured
+in the **lifetime of a label** — and the label is usually accurate when written,
+which is what makes it trustworthy right up until it is not.
+
+Worked instance: a blob was tagged `draft/amendment-15-links` at the moment an
+amendment was split in two. The tag was created by one agent, used by both, and
+verified three times — it resolves, it recovers the expected line count, its
+hash matches. Every one of those checks confirmed the object was **INTACT**.
+None asked what it **CONTAINED**, because the name answered that and the name
+was true at save time. Seven review rounds later the blob still held the
+*pre-review* form of content that had since been ratified in a much-changed
+state. Applying it as a "mechanical rebase" — the phrase both agents kept using,
+accurate when coined — would have silently reverted every P0 and P1 from those
+seven rounds while looking like a routine port of held work.
+
+**Integrity checks confirm an object has not changed. They say nothing about
+whether it is still the right object.** A checksum cannot detect that the world
+moved.
+
+A sweep naming only the first species passes the other two cleanly. The general
 form: *a check separated from the action it authorises is only as good as the
 interval between them; for anything under concurrent write the interval must be
 zero or the check must be REPEATED AFTER the action.* **Verify-then-act is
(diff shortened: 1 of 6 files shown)


zoolok17:
Closing on operator approval of the 2026-09-17 backlog triage: the native-work-spine RFC effort is paused. The document stays on branch feature/native-work-spine for reference.

PR #34, opened 2026-07-19
test: deterministic zero-spend enforcement canary (permanent E2E coverage)

zoolok17:
Adds a hermetic, provider-free, deterministic end-to-end canary for the detection-grade owed-action enforcement — permanent coverage for the print-not-run fail-open class and off-by-default behavior.

**Test-only** (`tests/fixtures/enforcement_canary_stub.py` + `tests/test_enforcement_canary.py`, 525 lines; no product code touched).

Validated by codex-2 against re-merged master (`12c7d21`), rebased SHA `ae16da1`:
- Scenario A (compliant): admit→reserve→1 canonical reply→SATISFIED.
- Scenario B (print-not-run): cap=2 respected, `DELIVERY_FAILED` emitted + requester-visible, breaker trips.
- Scenario C (inactive/no policy): 0 dispatches, ordinary legacy turn (off-by-default).
- Oracle sensitivity: removing ACTIVE policy fails at the admission oracle; restoring passes — enforcement-sensitive, not tautological.
- Full pytest 2959 passed (Py3.14); focused green on 3.10 + 3.14; ruff clean.
- **Actual provider/model/OVH calls: 0. Billed spend: €0** (paid-dispatch counts are internal gate accounting via a local Python stub).

### DIFF
--- a/tests/fixtures/enforcement_canary_stub.py
+++ b/tests/fixtures/enforcement_canary_stub.py
@@ -1,0 +1,208 @@
+"""Deterministic subprocess used by the detection-grade enforcement canary.
+
+The wrapper launches this file as if it were the Codex CLI.  It emits the small
+JSONL event subset consumed by the real Codex adapter and either executes the
+exact owed-action transport from the prompt or merely prints prose.
+"""
+
+from __future__ import annotations
+
+import hashlib
+import json
+import os
+from pathlib import Path
+import subprocess  # nosec B404 - the canary executes the gate-pinned argv only
+import sys
+
+
+CONTROL_ENV = "AGENTTALK_ENFORCEMENT_CANARY_CONTROL"
+OWED_MARKER = "== OWED ACTION TRANSPORT =="
+BUS_TIMEOUT_SECONDS = 15.0
+
+
+def _emit(value: dict) -> None:
+    print(json.dumps(value, ensure_ascii=False), flush=True)
+
+
+def _owed_action(prompt: str) -> dict | None:
+    if OWED_MARKER not in prompt:
+        return None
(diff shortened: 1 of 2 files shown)


zoolok17:
Closing as **superseded**, with the delta captured rather than dropped.

The feature this PR was built for (task #34, the deterministic zero-spend enforcement canary)
shipped in master as `tests/test_stub_agent_canary.py`, whose module docstring names task #34
explicitly. It drives real turns through `wrapper.run.make_drive` with the default spawner and a
real stub-CLI subprocess, which is what this PR set out to prove.

Why close rather than rebase:
- 8 days stale against a fast-moving tree.
- CI here is the **retired pre-dev-gate workflow** (`pytest (3.x on ubuntu-latest)`), every leg red.
  The repo now gates on `dev-gate` legs plus CodeQL, so these results say nothing about the
  current tree either way.
- An unaddressed connector P1 at `tests/test_enforcement_canary.py:189`.

**This was not a pure duplicate, and I am not pretending it was.** Three scenarios here have no
counterpart in master's canary:

- `test_compliant_stub_executes_reserved_reply_once_without_penalty`
- `test_print_not_run_is_recovered_then_becomes_visible_failed_delivery`
- `test_unconfigured_stub_keeps_legacy_wrapper_behavior_inert`

The last is the one I most want back: a negative control proving the enforcement machinery stays
inert when unconfigured is what gives the positive cases their meaning. The exactly-once assertion
in the first is also not made anywhere in master today.

Those three are tracked to be re-derived against master's canary and current APIs, which is cheaper
than rebasing this branch plus clearing the P1 on a workflow that no longer runs. The branch stays
readable at its head for reference.

Issue #16, opened 2026-06-05
Broadcast fan-out: preflight + batch manifest + partial-failure surfacing


zoolok17:
**Friction:** fan-out is a bare per-recipient loop; a mid-loop failure delivers recipients 0..k-1 with no report (the summary line never prints) and no rollback. Carried from ROADMAP ("fan-out is not transactional"); becomes production-relevant as role audiences expand fan-out usage.

**Proposed:** preflight all recipients before the first write; write a batch manifest under a shared batch id; on failure, print exactly who received the message and who did not; `status`/`threads` make partial fan-out visible. No true rollback (multi-file atomicity isn't available on a local FS) — explicit partial-failure surfacing instead of pretending atomicity.

---
From the 0.11.0 build notes + 2026-06-05 reprioritization; see ROADMAP.md. Same release as role-scoped audiences.

zoolok17:
Shipped in [v0.15.0](https://github.com/zoolok17/agenttalk/releases/tag/v0.15.0): preflight + frozen batch facts on every copy, exit-5 delivered/missed manifest on partial failure, one-command recovery via `broadcast --resume <bid>` (born in cross-review), and `incomplete fan-out` status warnings until resumed or rescinded. No rollback pretense - honesty instead, per the design consult.

PR #49, opened 2026-07-20
feat(dev-gate): hermetic SHA-bound gate command (#50) — ADOPTED from crashed builder, hold merge

zoolok17:
## #50 — hermetic SHA-bound dev-gate command

Implements the approved design: `agenttalk dev-gate` (manifest-driven, source+wheel, isolated temps, machine-readable SHA-bound evidence; CI invokes the same command). Addresses all three lead fold-points — manifest separates `local` (3.10/3.14) from `ci` (3.10–3.13) python matrices; pip-audit/semgrep/zizmor folded in as checks; packaging/wheel-contract assertions retained. Rewires `tests.yml` + `security.yml` to call `dev-gate`.

### ⚠️ Provenance — ADOPTED, held for operator merge
codex-2 built this to `06e742e` (full arc: runner → evidence trust gaps → tool/wheel/git-fixture isolation) then its **wrapper crashed mid-turn** (context-bloat, task #16) **before sending review-ready** — so there is no builder-supplied final evidence artifact or gate-summary. The lead adopted the committed work. **Do not merge on green CI alone:** this rewrites CI workflows (high blast radius) and the builder never self-declared done. Wants a human review before merge — especially the `tests.yml`/`security.yml` changes.

- Bus review requested from claude-remediation-reviewer (independent pass).
- all-OS CI running (also self-tests the reworked workflows).
- Lead: focused red-flag scan done; full review deferred to fresh eyes.

### DIFF
--- a/.github/workflows/security.yml
+++ b/.github/workflows/security.yml
@@ -6,137 +6,19 @@ on:
   pull_request:
     branches: [master]
   schedule:
-    # Weekly catch-up so newly disclosed CVEs in dev deps surface
-    # even when the repo is quiet.
-    - cron: '17 9 * * 1'  # Monday 09:17 UTC
+    - cron: '17 9 * * 1'
   workflow_dispatch:
 
-# Default workflow permissions: read-only. Jobs that need write
-# access (e.g. CodeQL SARIF upload) grant themselves the minimum
-# required at the job level. zizmor's `excessive-permissions`
-# audit prefers this pattern.
 permissions:
   contents: read
 
-# Third-party actions are hash-pinned (per zizmor `unpinned-uses`
-# / OWASP CI/CD recommendations). The `# vX` trailing comment is
-# human-readable and dependabot-friendly — dependabot auto-bumps
-# the SHA while keeping the comment in sync.
-#
-# Every actions/checkout sets `persist-credentials: false` (zizmor
-# `artipacked`): the default writes GITHUB_TOKEN to the local git
-# config, which can leak via uploaded artifacts. None of these jobs
-# push back to the repo, so the credential is just dead weight.
-
+# CodeQL is the sole CI-native exception declared in dev-gate.json. Every
+# CLI-runnable voting check lives in the committed dev-gate plan instead.
 jobs:
-
-  # --------------------------------------------------------------- ruff S
-  ruff:
-    name: ruff (lint + security S-rules)
-    runs-on: ubuntu-latest
-    steps:
-      - uses: actions/checkout@df4cb1c069e1874edd31b4311f1884172cec0e10  # v6.0.3
-        with:
-          persist-credentials: false
-      - uses: actions/setup-python@a309ff8b426b58ec0e2a45f0f869d46889d02405  # v6.2.0
-        with:
-          python-version: '3.12'
-      - name: Install ruff
-        run: pip install 'ruff>=0.6'
-      - name: Run ruff
-        run: ruff check src tests
-
-  # --------------------------------------------------------------- bandit
-  # Overlaps with ruff S-rules; we run both for defense-in-depth
-  # (each catches things the other misses or scores differently).
-  bandit:
-    name: bandit (security AST scan)
-    runs-on: ubuntu-latest
-    steps:
-      - uses: actions/checkout@df4cb1c069e1874edd31b4311f1884172cec0e10  # v6.0.3
-        with:
-          persist-credentials: false
-      - uses: actions/setup-python@a309ff8b426b58ec0e2a45f0f869d46889d02405  # v6.2.0
-        with:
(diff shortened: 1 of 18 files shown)


zoolok17:
**Lead diagnosis of the CI failures (adopted PR; builder codex-2 crashed pre-review-ready).** The gate correctly **fail-closed (BLOCK)** rather than false-passing — good. Two real, distinct issues, both of which dev-gate surfaces *because it is stricter than the current editable-install CI*:

**1. Windows — 11 wheel-mode pytest failures (all Windows legs).** `pytest-wheel-py310` etc.: `11 failed, 3059 passed` against the **built wheel** (source mode passes). Failures:
- `test_powershell_functional.py`: 3 (real_core parses/runs, direct-pwsh→python claim chain, override baked fallback)
- `test_supervisor.py`: 7 (`test_generated_ps1_*`: malformed-config, hot-add, holds-poll, two-polls-no-dup [legacy-direct & wrapped], quiet-suppresses-warning, quiet-suppresses-relaunch)
- `test_coordination.py::test_scoped_wait_replacement_receives_later_reply`: 1

Only Windows fails because these run real `pwsh` (present on Windows; skipped on linux/mac). They pass editable but fail on the wheel → **the built wheel is very likely missing PowerShell template/harness package-data** (pyproject data-inclusion gap), or those tests need source-tree access and should be scoped source-only. Either way it's a REAL find that editable-only CI has been masking — the point of wheel-mode gating. `wheel-contract` currently only asserts the web_static assets; it should also assert the pwsh data files.

**2. linux/3.12 static leg — semgrep BLOCK (exit 1, findings).** Reconcile dev-gate's semgrep handling with how the current `security.yml` semgrep job treats findings (baseline / `--error` / tolerance). If current CI tolerates these, dev-gate is stricter; decide which is intended.

**Status:** HELD for operator. Not fixed overnight (builder crashed; substantial). Next: decide (a) fix wheel packaging vs. scope pwsh/supervisor tests source-only, (b) semgrep tolerance, then re-gate. macOS + linux 3.10/3.11/3.13 legs passed clean.

zoolok17:
**Correction to my prior diagnosis (checked, disproved my own hypothesis).** There are NO PowerShell *data* files to package — pwsh support is all code (`powershell_host.py`) + runtime-generated `.ps1`, and hatchling already packages non-.py assets. So the 11 Windows wheel-mode failures are **not** a missing-data-file packaging gap. Root cause is still open: the pwsh-functional + supervisor `generated_ps1` tests fail against the **wheel-installed** agenttalk on Windows specifically (pass in source mode; skipped on linux/mac where pwsh is absent). Likely a test↔wheel-install interaction — e.g. a generated supervisor.ps1 or a test fixture resolving `agenttalk` / `python -m agenttalk` against site-packages vs the source tree, or a test that assumes an editable layout. The fix is 'investigate wheel-mode pwsh test behavior', NOT 'add package-data'. Needs fresh investigation (builder crashed). Semgrep item stands.

zoolok17:
**ROOT CAUSE (definitive, with proof) — the Windows wheel-mode failures are a test-harness artifact, NOT a product bug.**

Reproduced locally against the built wheel, then isolated the mechanism:
```
wheel-test venv python:  Popen.pid=24416   child os.getpid()=20208   MISMATCH
real pinned interpreter: Popen.pid=4044    child os.getpid()=4044    MATCH
```
The wheel-test **venv `python.exe` is a redirect launcher** that re-execs the real interpreter as a different-pid child. The 11 failing tests (test_coordination marker, test_supervisor `generated_ps1` pid tracking, test_powershell_functional) all assert that a spawned process's recorded pid equals `Popen.pid`. Under a real interpreter those match (source/editable CI → pass); under the venv launcher they don't (dev-gate wheel-mode → the spawned agenttalk writes a valid marker but with the child pid, so the parent's `marker.pid == p1.pid` poll times out). Confirmed the subprocess itself runs cleanly (no stderr; marker written correctly) and agenttalk imports in ~94ms (not a slow-import or packaging issue).

**Implications:**
- Product is exonerated: not an agenttalk bug, not a wheel-packaging gap; **shipped wheels are unaffected** (this only bites tests that assert pid-identity when spawning through a venv launcher).
- Fix belongs in the dev-gate HARNESS: build the wheel-test venv so `python.exe` is not a redirect launcher (e.g. `python -m venv --copies`, or resolve/point tests at the venv's real base interpreter so `sys.executable` is the actual exe). Then the 11 pass (they pass on a real interpreter).
- The linux/3.12 **semgrep** block is separate and still needs the tolerance/baseline reconciliation.

So: dev-gate's stricter wheel-mode surfaced a real venv-launcher pid-identity assumption in the test suite — good signal — and the fix is a small harness change, not a product change.

PR #71, opened 2026-07-22
fix(#62): transient gateway hold retries-through so a held worker self-heals

zoolok17:
## #62 — a held gateway now self-heals instead of sticky-parking

The last piece of the Qwen-reliability arc (after #58 made the wedge visible, #63 stopped the doomed held-mint).

**The gap:** after #63, a held gateway cleanly *refuses* to mint a child turn (`LedgerHold`) — but the wrapper classified that refusal as `config_blocked`, which **sticky-parks the inbox head and never re-drives, even after the operator clears the hold.** So a worker blocked only because the gateway was temporarily held stayed parked until a manual reset. This was Fable's finding on #70 (I'd wrongly claimed #63 auto-recovered).

**Fix:** a held / unresolved / clock-rollback gateway state is **transient and operator-resolvable**, so classify it **INFRA** (retry-through under backoff, never dead-lettered) — the worker resumes automatically the instant the hold clears. Two coordinated paths, both required (fixing only one still sticky-parks via the other):

1. **Mint-time:** `_GatewayChildCapUnavailable` carries a `transient` flag, set iff the underlying error is a `LedgerHold` (type-based — `ChildTurnCapExceeded` is a `LedgerBlocked`, **not** a `LedgerHold`, so a permanently-capped message never takes this path). The spawner emits an explicit `gateway_transient_hold` signal → `_classify_drive_failure` → `CLASS_INFRA`. No child spawned, no spend — the retry is cheap and pre-model.
2. **Mid-turn:** a 503 `ATGW_LEDGER_BLOCKED` now classifies INFRA instead of config_blocked. This deliberately reverses the prior "ledger block is terminal" test assumption — **that assumption *was* the sticky-park bug.**

**The trap I did NOT trip:** a per-message child-turn cap (calls/cost/**wall-time**, `ATGW_CHILD_TURN_CAP_EXCEEDED`) stays `config_blocked` — a capped/expired turn can never re-mint the same `message_id`, so retrying it would spin forever and never dead-letter. Policy block (trial cutoff / model mismatch) also stays terminal.

**Verification.** 293 gateway+wrapper tests green, ruff + bandit clean. New/updated tests: `gateway_transient_hold` → INFRA (backend-agnostic); `ATGW_LEDGER_BLOCKED` → INFRA (updated the test that encoded the old terminal assumption); cap + policy still config_blocked; exception carries the flag.

**Review focus (cross-model welcome):** the `isinstance(exc, LedgerHold)` discriminator (does it correctly exclude every permanent-cap case?); whether reclassifying `ATGW_LEDGER_BLOCKED` as retryable can spin on a *non-self-clearing* LedgerBlocked (e.g. DB-open contention vs corruption); and the INFRA retry/escalate/never-DL semantics for an overnight hold.

**Deliberately deferred (documented):** the general `config_blocked` re-probe (a genuine local exec-denial later fixed) — a genuine local denial needs a human fix, so escalate+park is appropriate. This PR closes the high-value case: the hold.

Closes #62 (hold self-heal). Residuals (reap stuck-open turns; mid-turn-hold wall-clock; multi-agent `_unresolved` false-park) remain tracked notes.

### DIFF
--- a/src/agenttalk/wrapper/run.py
+++ b/src/agenttalk/wrapper/run.py
@@ -59,7 +59,19 @@ _BENIGN_PIPE_TEARDOWN_ERRNOS = {errno.EINVAL, errno.EPIPE}
 
 
 class _GatewayChildCapUnavailable(RuntimeError):
-    """The wrapper could not durably issue a scoped paid-child credential."""
+    """The wrapper could not durably issue a scoped paid-child credential.
+
+    ``transient`` marks an operator-resolvable gateway HOLD (a durable accounting
+    hold, an unresolved prior attempt, or a clock rollback needing reconciliation -
+    i.e. any ``LedgerHold``) as opposed to a durable local/config denial. A transient
+    hold is classified INFRA (retry-through, never dead-lettered) so a worker blocked
+    only because the gateway is held self-heals when the hold clears, instead of
+    sticky-parking as config_blocked forever (#62). A NON-transient cause (missing
+    message id, ChildTurnCapExceeded, a local exec denial) stays config_blocked."""
+
+    def __init__(self, message: str, *, transient: bool = False) -> None:
+        super().__init__(message)
+        self.transient = transient
 
 # Legacy infra-like text markers. These are low-confidence hints only; they deliberately
 # no longer classify known_global_infra without a structured CLI status/rate-limit fact.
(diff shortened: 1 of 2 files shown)


zoolok17:
Cross-model (Fable) review found the naive INFRA reclassify only half-delivers and regresses the overnight-hold workflow — converting to DRAFT, not merging. Confirmed: (1) 'never dead-lettered' is false — CLASS_INFRA dead-letters at the 4h infra-exhaustion backstop, so a message queued during a long hold is DLQ'd rather than preserved; (2) mid-turn holds >300s still park after clear (expired turn can't re-mint); (3) tests don't exercise the real spawner→LedgerHold path (except-ordering unpinned); plus the INFRA path withholds the heartbeat by design → supervisor restart-storm all night under a hold. Root cause: a deliberate operator HOLD is not a generic infra OUTAGE. Correct fix = a distinct 'blocked-but-alive' state (heartbeat fresh, cheap re-probe, never counts toward exhaustion, resume on clear). Redesign pending operator steer; see #62.

zoolok17:
Superseded by #72. The CLASS_INFRA approach here regressed the overnight workflow (dead-letters at the 4h infra-exhaustion backstop + withholds the heartbeat → supervisor restart-storm), as Fable review found. #72 introduces a dedicated CLASS_GATEWAY_HELD (park without persisting an attempt → never dead-lettered → self-heals on clear) scoped to the mint-time hold, with a one-shot escalation for autonomous holds. Closing this draft.

Issue #35, opened 2026-07-19
deadman config in supervisor.json is silently inert (actually read from .agenttalk/config.json)


zoolok17:
## Summary

`supervisor.json` advertises a `deadman` block that **nothing reads**. The deadman's actual config lives in `.agenttalk/config.json`. Tuning the knob where the generated template shows it is a **silent no-op**, so the operator believes a safety mechanism is tuned when it is not.

## The split

- `src/agenttalk/deadman.py:88` — `cfg = store.load_config()` → reads **`.agenttalk/config.json`**
- `src/agenttalk/supervisor.py:4694` — the supervisor.json template **agenttalk itself generates** contains:
  ```jsonc
  "deadman": { "mail_age_slo_seconds": 900, "alarm_unread_response": false },
  ```
- `src/agenttalk/store.py:976` — the same defaults, in the store config, which is the surface that is actually read.

Two writers, one reader, and the writer an operator is most likely to reach for is the wrong one.

## Reproduction (verified live)

```
# 1. set the knob in supervisor.json, where the generated template documents it
"deadman": { "mail_age_slo_seconds": 2700, "alarm_unread_response": true }

# 2. ask the deadman what it is using
$ agenttalk deadman --json
  threshold_seconds     : 900.0     <-- ignored
  alarm_unread_response : False     <-- ignored

# 3. move the identical block to .agenttalk/config.json
$ agenttalk deadman --json
  threshold_seconds     : 2700.0    <-- applied
  alarm_unread_response : True      <-- applied
```

## Why this rates higher than a docs nit

1. **It fails silently.** No unknown-key warning, no validation error. `_deadman_config` validates the block's *contents* carefully (type check, positive-number check, bool check) — but only after it has already read the right file. So a well-formed block in the wrong file receives the full strictness of no checking at all. The failure mode is indistinguishable from "configured correctly".
2. **The affected mechanism is the one that catches a non-listening lead.** A knob that appears set but is not is worse than no knob: it converts an unmonitored gap into a monitored-*looking* one. This surfaced here only because I checked whether my own tuning had taken effect. Nothing would have told me otherwise.
3. **The defaults are the ones most likely to need tuning.** Shipping `900` / `alarm_unread_response: false` is defensible as a default, but both wanted changing on first real contact (see below) — which puts this trap directly on the common path, not in a corner.

## Calibration note (separate finding, non-blocking)

Independent of the config split, the two buckets appear tuned backwards for a lead-with-builders topology:

[...]

PR #78, opened 2026-07-24
feat(#60 inc-3): DoD coverage dimension — producer + floor + parser

zoolok17:
## Summary

This completes #60 increment 3 by connecting configured coverage scans to the
Definition-of-Done coverage floor without adding a runtime parser dependency.

- Parses recognized coverage.py and pytest-cov terminal summaries from stdout.
- Publishes the measured percentage to exactly one finite producer gate,
  `coverage:change`, `coverage:release`, or `coverage:deep`, unless that gate
  holds an active persisted operator waiver.
- Lets each close scope explicitly select a `coverage:<profile>` gate plus its
  minimum percentage and optional maximum evidence age.
- Fails closed on missing, malformed, stale, low, wrong-profile, or
  wrong-revision evidence.
- Preserves an active operator waiver when a later automated scan completes.

## Coverage command contract

Stdout is the sole coverage-evidence channel. The parser does not consume
`coverage.json`, `coverage.xml`, or any other report file.

The scanner recognizes exactly two canonical root report paths:

- `coverage.xml`
- `coverage.json`

While holding the per-root coverage lock, it classifies both paths before
running the command. If either name exists as any filesystem object, including
a regular file, symlink, or directory, the scanner names every conflict, records
a red automated result, and does not run the command. An active persisted
operator waiver remains unchanged. The scanner never quarantines, restores,
consumes, or deletes either path.

After the command returns, the scanner checks both names again before accepting
stdout evidence. When postflight observes a path created or replaced during the
run, agenttalk does not read its contents, move it, or remove it; the automated
result is red unless preserving an active operator waiver, and the error names
the path. Branch-local legacy coverage recovery residue is also left in place
and reported with its root and backup paths for manual recovery; legacy marker
fields are never treated as proof of authorship.

This is an agenttalk-orchestration guarantee, not filesystem isolation for the
configured command. The subprocess and its descendants can create or modify
paths while running. Process containment or an owned-output protocol remains
#107.

This refusal model replaces the earlier custody model. A pathname, phase,
revision, receipt, or digest cannot prove who wrote the exact object currently
at a shared root path, so guessing ownership would reintroduce silent operator
data loss.

## Exact boundary and residual risk

The protected and discovered path class is exactly `coverage.xml` and
`coverage.json`. An arbitrary configured command can write another path, which
agenttalk neither discovers, parses, nor cleans. On case-sensitive filesystems,
case variants are outside the class.

A surviving producer descendant can create a canonical report after postflight.
The next scan then refuses until the named file receives manual cleanup, even
when that late report came from agenttalk's own prior command. This deliberate
false-DOWN remains until #107 provides process containment or an owned-output
protocol; it is safer than deleting a file whose author cannot be proven.
Any coverage command that spawns subprocesses can plausibly trigger this on an
ordinary setup; until #107, the operator remedy is to inspect and manually clean
up the named path.

Executed proof for Windows symlink/reparse refusal remains unconfirmed: the
local symlink cases skip with `WinError 1314`, and CI does not explicitly grant
symlink-creation privilege.

Stdout evidence is rejected when its UTF-8 representation exceeds 16 MiB, and
numeric parsing is fail-closed for non-finite or out-of-range percentages. The
size bound applies when parsing:
`capture_output=True` still buffers the complete subprocess streams first.
Capture-time bounding is tracked separately as #106.

## Stdout evidence grammar

Coverage subprocess stdout is captured as BYTES and decoded as strict UTF-8 with
an optional BOM, so bare `
` is preserved rather than translated to `
` by
text-mode capture. Only coverage.py `TOTAL` and pytest-cov `Total coverage:`
summaries are recognized; bounded SGR colour decoration is accepted, and the
final recognized summary across both formats wins.

Refused, all failing red: bare-CR progress rewrites, non-SGR or malformed
terminal controls, stderr-only summaries, locale-comma decimals, invalid UTF-8,
malformed or out-of-range percentages, and any command exiting nonzero.

Only an ACTIVE, VALID operator waiver suppresses automated writes. An expired or
invalid waiver is replaced by fresh evidence, or invalidated when no fresh
measurement exists, so the gate cannot become permanently unsatisfiable.

agenttalk cannot detect an inner failure hidden by a wrapper that exits zero;
whichever recognized summary prints last is used. Such wrappers must propagate
failures and print the intended aggregate last.

## Definition-of-Done integration

Coverage scan profiles and close scopes are separate policy axes. A close scope
must name one of the producible `coverage:<profile>` gates and configure a
finite `min_percent`; it may also configure `max_age_days`.

The resolver requires automation-CI evidence from the selected producer profile,
bound to the close revision. It reports distinct holds for missing, stale, and
below-floor coverage. A gate waiver does not silently substitute for the
revision-bound DoD evidence contract.

## Verification

The regression suite covers:

- stdout parsing boundaries and malformed producer output;
- pre-existing regular files, symlinks, and directories for both canonical
  report names;
- byte-identical preservation and proof that a refused command never starts;
- a canonical path created by an external writer after clean preflight;
- clean-tree stdout evidence, gate publication, waiver preservation, and
  non-waived gate updates;
- legacy recovery residue refusal without restoration or deletion;
- concurrent coverage scans without cross-claiming evidence;
- coverage-floor resolution for profile, revision, age, and percentage.

Review status: the round-10 custody removal and the round-11 stdout-evidence
fixes are both independently reviewed and approved, each verified by direct code
reading plus executed reproduction rather than by accepting the change
description.


zoolok17:
## Round 6 → 7: operator-waiver clobber closed — `d1cb60a` → `498f702`

An independent review reached a **blocking** finding five prior rounds had not: the coverage gate's compare-and-swap protected a narrower thing than "the gate".

`_invalidate_stale_coverage_gate` (the no-fresh-measurement finalizer) was correctly CASed **and** checked `status == "waived"`. `_emit_coverage_gate` — the write that records the scan's actual green/red result — took no expected value, performed no comparison, and had **no waived check at all**. Nothing upstream skipped the scan when the target gate was already waived. So, with no race required and in either timing order:

```
operator waives coverage:release
  → any later coverage scan for that profile runs to completion
  → _emit_coverage_gate silently overwrites the waiver with a fresh red or green
```

Direction: **fail-OPEN of the operator's own decision** — false-GREEN when the clobbering scan is green, and the deliberate override is erased without the operator seeing it change. The existing revision-binding check in `close.py` does *not* reach this case: a waiver-clobber followed by a same-revision green passes it cleanly, because there is no mismatch to catch — only an erased decision.

### Fixed

`_emit_coverage_gate` now takes the root config lock, reloads the persisted target gate **inside** that lock, and returns without writing when its status is `waived`. Check and write are one critical section. Non-waived gates still update normally; same-revision non-waived reruns remain last-write-wins, which the review explicitly scoped as inherent rather than a defect.

### The channel class, enumerated

The reviewer's rule for this round was to state the *class*, not the instance. Every production path that can persist a coverage gate status:

| path | treatment |
|---|---|
| `_emit_coverage_gate` (fresh completed measurement) | config-locked, persisted-state waiver guard |
| `_invalidate_stale_coverage_gate` (no fresh measurement) | snapshot + config-lock CAS, waiver-aware, duplicate stale-red suppressed |
| `gate waive` → `gates.waive_gate` | exempt — this operation *creates* the protected waiver |
| `gate set` → `gates.set_gate` | exempt — explicit typed operator replacement; generic `set` cannot mint status `waived` or source `operator_waiver` |

`gates.set_gate` / `gates.waive_gate` stay policy-free lock-free primitives; every production call site supplies the lock and policy. Named residual: a direct third-party library caller can bypass CLI serialization — not an automated assurance path.

### Two siblings, same shape

- **`coverage_parse.py`**: `_from_json` enforced `MAX_COVERAGE_ARTIFACT_BYTES`; `_from_stdout` enforced nothing before regex scanning, though stdout comes from the same untrusted producer subprocess. Now bounded for str and bytes, measured by **encoded byte length** rather than character count. Disclosed residual: `capture_output=True` still lets the OS buffer before the parser bound applies — this bounds parser/regex work, not capture-time memory.
- **`assurance.py`**: a failed generated-artifact removal only blocked transaction completion for artifacts that *had* a backup; a freshly-created artifact with no backup closed the transaction as "done" while leaving an orphan. Now completion-blocking regardless of backup presence, for both JSON and XML.

### Evidence

Seven new test functions (11 cases) against the reviewed baseline `d1cb60a`: **7 failed, 4 passed** — the four passes being the controls. Same selector at `498f702`: **11 passed**. Affected module suite: `96 passed, 2 skipped`.

Repository-wide: `2 failed, 3831 passed, 21 skipped`. Both failures are `tests/test_ovh_gateway_service.py` probing fixed loopback ports 4000/4001, which had pre-existing listeners (PIDs from 2026-07-22 — the operator's live gateway, correctly left running). No other test failed. Unrelated to this diff; it is another instance of the fixed-resource family already tracked.

`ruff` clean, `bandit` 0 findings across 69,794 lines, `semgrep` 0 blocking, `compileall` 0, `git diff --check` 0.

Current master (`d7fe1b9`) merged in first. A confirming review of this fold is running, including an explicit question about what the new guard now *refuses* that it should not.

zoolok17:
## Merge gate held at `498f702` — two fresh connector findings, one P1

CI is 15/15 green on the reviewed head and round 7 is approved, so this is not a CI or review problem. The connector gate caught a path no lens covered.

### P1 — `assurance.py:1485`, held for assessment

> When a coverage command times out after spawning worker or grandchild processes, `_run_external` terminates and waits for only the direct subprocess, then this path immediately removes generated reports and restores the quarantined originals. A surviving descendant can subsequently write to `coverage.json` or `coverage.xml`, overwriting the restored operator-owned report after the transaction lock has been released.

If it holds, it defeats the transaction's purpose: the mechanism exists so an operator's own report is never destroyed, and this destroys it *after* the restore is declared complete and the lock is dropped. That is data custody, not bookkeeping.

Sent for a **premise check before any build**, because the finding stacks four independent claims and is only real if all of them hold: that the timeout path terminates only the direct child; that the configured coverage command can actually spawn descendants which outlive it; that such a descendant would still write a report at that point rather than dying mid-run; and that the restore genuinely lands outside the lock. A refutation with evidence is an acceptable answer.

If it does hold, the fix has to be cross-platform — this layer was already bitten once by a Windows-only implementation (#35) — and must answer its own direction-of-failure question, since a process-tree kill that cannot prove ownership would trade "the report may be overwritten" for "we killed something we did not own."

### P2 — `assurance.py:1455`, disposed as an already-accepted residual

`capture_output=True` buffers complete streams before the 16 MiB evidence bound applies, so a malfunctioning command can exhaust memory before the fail-closed bound is reached.

This is the residual **disclosed by the builder and explicitly examined and accepted by the reviewer** in round 7. The finding that round closed was that `coverage_parse.py` regex-scanned an unbounded string — a parser-internal concern, now bounded for str and bytes by encoded byte length before either regex runs. The remaining exposure is one layer up in how the subprocess is invoked, which a pure parser structurally cannot bound; closing it means streaming or capped capture, a materially larger change than what was asked for.

Filed as its own task rather than folded, so the disposition is durable instead of being re-litigated every round. Deliberately kept out of round 8 so the P1 is isolated.

### The three older comments on this PR

Stale — anchored to earlier commits and addressed in rounds 4 through 7 (the never-raises contract as a class, the crash-recovery report deletion, the waiver clobber and its two siblings).

PR #83, opened 2026-07-24
feat(#71 B1): install checkpoint PreCompact/SessionStart hooks via the hook installer

zoolok17:
#71 fast-follow B1: extend the existing `install_activity_hook` path (supervisor.py) so it idempotently installs/upgrades the Claude `PreCompact` (checkpoint save) and `SessionStart/compact` (checkpoint resume) hooks alongside the heartbeat hook — no hand-edited settings.

**Rollout guard (the key detail):** each checkpoint command falls back through `|| agenttalk heartbeat --hook` (same `--fallback-for` identity), NOT `|| exit 0` — because pwsh treats `|| exit 0` as an unresolved command in this chain context. So on an install predating the `checkpoint` subcommand, the hook exits rc 0 via the always-present, silent, fail-soft heartbeat and never blocks a compaction or contaminates SessionStart stdout. `||` is valid on POSIX sh, cmd.exe, and pwsh 7.

Tests: fresh/existing settings, matcher-group shape, idempotence, legacy upgrade/dedupe, fallback identity, cmd/pwsh/POSIX-sh old-subcommand behavior, `--codex` heartbeat-only boundary. 260 supervisor + focused install suites green; ruff/bandit clean; semgrep 0 findings (dev-gate config). `.codex/hooks.json` stays heartbeat-only (B2 untouched). Requires #80's checkpoint command in the runtime (ships same release).


zoolok17:
Merging with ONE consciously-accepted, documented residual — recording the reasoning rather than merging past an unread finding.

**Fresh connector P2 at `supervisor.py:6699` (orig=57b8854): "Use one fallback identity for both checkpoint events."** When the installer upgrades a *partial* identity-bound configuration (e.g. only `PreCompact` currently carries `--fallback-for <agent>`), each event discovers its fallback independently — so `save` keeps the bound identity while `SessionStart` is installed neutral. In a liaison window with no `AGENTTALK_SELF`, save then writes that agent checkpoint while resume cannot resolve an identity and emits empty context, even though the installer reports both hooks as authoritative. That is a real instance of the same silent-false-assurance class as F2 and the `type` P2 already fixed here.

**Why it is being accepted for v0.79.0 rather than blocking it:**
- **Not reachable on a first install.** The trigger requires pre-existing checkpoint hook entries in a partial identity state. The feature has never shipped, so no such entries exist anywhere yet; a fresh install creates both checkpoint hooks together and inherits one fallback from the heartbeat entry (`supervisor.py:6813`).
- It becomes reachable only after a partial state arises later (e.g. a reported `skipped (malformed ...)` event that is then repaired and re-run) — and those skips are now reported honestly rather than silently, thanks to the F2 fix in this PR.
- The release has been gated on this PR all night and is the operator explicit ask; the correct trade is to ship and fix, not to hold an installable release for a defect that cannot fire on the path the operator will take.

**Committed follow-up:** a fix is dispatched immediately as the first item for v0.79.1 — resolve ONE fallback identity across both checkpoint events before merging them, with a regression covering the partial-config upgrade (only-PreCompact-bound, only-SessionStart-bound, and both-neutral) asserting a single consistent identity and non-empty resume context in a no-`AGENTTALK_SELF` window.

Gate: 15/15 green at `57b8854`; the other connector comment (`:6707`, orig=7bfcc70) is the `type` P2 already fixed in this PR.

PR #92, opened 2026-07-28
fix(#93): pin the control-plane tool runtime off the checkout (tool_runtime_python)

zoolok17:
Adds `tool_runtime_python`: a supervisor-config field that pins the **control-plane** runtime independently of whether the bus store happens to live inside an agenttalk checkout.

Unset (the default) is byte-identical to today. That is also the rollback: remove the field, refresh, restart.

## The defect

`supervisor.py` renders `$SrcOnPyPath` from `src_is_checkout = (store.root/"src"/"agenttalk"/"__init__.py").exists()`. Because this project's bus store lives inside its own checkout, it is always true, so the wrapped-agent launch path does:

```powershell
$applied = @{ AGENTTALK_ROOT = $Root; AGENTTALK_PY = $AgenttalkPython }
if ($SrcOnPyPath) { $applied['PYTHONPATH'] = (Join-Path $Root 'src') + ';' + $env:PYTHONPATH }
```

So **every bus/control command every wrapped agent runs imports the live editable dev tree** — whatever half-edited state the working tree is in at that moment.

The intent behind `src_is_checkout` is defensible: *"your store lives in a checkout, so you are developing agenttalk — run the checkout's code."* That is right for code **under test** and wrong for the **tool**. One flag was serving two purposes that need opposite answers.

## Both doors, because one is not enough

Repointing the interpreter alone would have looked fixed and changed nothing. Measured, same interpreter with a non-editable agenttalk 0.79.0 in its site-packages:

```
PYTHONPATH unset        -> ...site-packages\agenttalk    0.79.0
PYTHONPATH=<repo>\src   -> <repo>\src\agenttalk          0.79.1
```

PYTHONPATH wins. So when the field is set, generation renders the pinned `$AgenttalkPython` **and** `$SrcOnPyPath = $false` **and** a `$ToolRuntimePinned` marker; the supervisor makes the pin authoritative *after* per-agent env application and clears `PYTHONPATH` (so an agent's own `env` block cannot reintroduce it); and the wrapper's computed child environment restores the same pin and strips inherited `PYTHONPATH`.

`bin/agenttalk.cmd` follows the tool runtime too — it is the supervisor's *other* control-plane door, so leaving it conditional would have reopened the hole from the side.

## Acceptance is a test, not an inspection

The load-bearing case executes the generated `Launch`, captures its effective environment, feeds the emitted runtime marker through `wrapper.run._child_env`, and actually runs:

```
<computed AGENTTALK_PY> -c "import agenttalk; print(agenttalk.__file__)"
```

asserting the result is under the isolated tool runtime and not the competing checkout. It also executes the generated `.cmd` shim under a **deliberately contaminated** ambient `PYTHONPATH` and a wrong `AGENTTALK_PYTHON`.

- Focused: `16 passed, 513 deselected`
- Affected suites (`test_powershell_artifacts`, `test_powershell_functional`, `test_supervisor`, `test_wrapper_loop`, `test_doctor`): **621 passed, 2 skipped**
- `ruff` clean, `git diff --check` clean, `pyproject.toml` untouched (`dependencies = []` stays)

Invalid values fail closed **before** any artifact write: non-string, blank, relative, missing, whitespace-padded, or Windows-shim-unsafe paths are rejected. Generated-artifact marker/body validation remains the only accepted writer, so a hand-edit is still refused.

## Rollout prerequisite found by execution — read before applying

A tool runtime built from the **previous** release cannot be used. Against a bundle generated by this commit:

```
CURRENT_GENERATOR_VALID=True
OLD_RUNTIME_VALIDATION=REJECTED
ArtifactValidationError: generated PowerShell artifacts are stale or mixed:
bin/agenttalk.cmd has a missing or duplicate baked Python fallback
```

Pointing `tool_runtime_python` at a v0.79.1-tag runtime would **wedge supervisor claim/validation**. The generator and the pinned runtime must agree on the artifact contract, and this commit changes that contract — so the runtime has to be built from source that *contains* this change. A matched pair was proven: `MATCHED_RUNTIME_VALID=True`.

Required order: **merge this → build the non-editable tool runtime from the merged source → set the field → refresh → restart, attended.**

That is a genuinely non-obvious chicken-and-egg property of introducing a pin: you cannot pin to the release *before* the one that adds pinning.

## Scope

**B1 is deliberately not built.** Binding an agent's *code under test* to *its own* checkout is a separate, lane-scoped mechanism; this global control-plane field is not it. Tracked in #93.

Built by `codex-2`. Verified clean before push: 7 files, +560/-17, `pyproject.toml` unchanged, `git status --porcelain` empty, `d5add08` an ancestor.


zoolok17:
**Correction to this PR body, from the review.** I wrote "Unset (the default) is byte-identical to today." That is **wrong** and it was my wording, not the author's.

Checked against the real pre-PR baseline (the PR's own test only compares unset vs explicit-null *under the new code*), unset mode renders `$ToolRuntimePinned = $false` plus five new gated blocks that did not exist before, and the generator marker hash changes. Every one was traced and confirmed **inert** — `$ToolRuntimePinned` is `$false`, so they never execute.

So the property that matters holds and is verified: **same env vars, same PYTHONPATH logic, same interpreter resolution — the rollback restores today's behaviour.** The correct phrasing is *functionally identical*, not *byte-identical*.

Also now blocking: the pin force-sets four keys and clears `PYTHONPATH`, but never `PYTHONHOME` or `PYTHONUSERBASE`. CPython puts user site-packages ahead of standard site-packages, so a leftover `pip install --user -e <checkout>` would let the pinned interpreter import the checkout with `PYTHONPATH` correctly empty. Being folded, with an executed differential.

zoolok17:
## Round: PS5.1 quote-stripping fix — `5c463f4` → `ce86fa4`

An independent lens executed the Python-3.11+ preflight guard under **Windows PowerShell 5.1** rather than accepting the structural-only claim, and found it broke the feature outright:

- The version-probe one-liner contained embedded double quotes (`"%d.%d"`).
- PS7 (`$PSNativeCommandArgumentPassing = Windows`) marshals that correctly; **Windows PowerShell 5.1 strips the quotes** before the argument reaches the child process.
- Consequence, traced through the real code: a genuine Python **3.14** interpreter crashed with `SyntaxError`, exited 1, `$rc -ne 0` fired, and `Preflight` returned `$false`. Because the corruption depends only on the one-liner's own quoting, **no pinned agent could launch under PS5.1 regardless of the configured interpreter's version** — with a diagnostic claiming the version was `unavailable` for an interpreter that was in fact 3.14.

Direction of failure was safe (false-DOWN, blocks rather than admits) but it was a total availability break of the feature on a shell this project treats as a live reality.

### The fix

One production line: the `-c` payload now uses `print(sys.version_info[:2])` — no embedded double quotes. Comparison and exit semantics unchanged (>= 3.11 exits 0; < 3.11 prints its real tuple and exits 1).

### New regression

A permanent test renders the real `Preflight`, creates a real venv interpreter with a minimal installed `agenttalk` module, and **executes that Preflight through both `pwsh` 7 and the canonical Windows PowerShell 5.1 executable**.

Failing-first on `5c463f4`: `[pwsh-7] PASSED`, `[windows-powershell-5.1] FAILED` with the exact `SyntaxError` above.

Real-interpreter matrix with the new payload:

| shell | pinned interpreter | rc | stdout |
|---|---|---|---|
| pwsh 7.6.3 | Python 3.14 | 0 | `(3, 14)` |
| Windows PowerShell 5.1 | Python 3.14 | 0 | `(3, 14)` |
| pwsh 7.6.3 | Python 3.10.11 | 1 | `(3, 10)` |
| Windows PowerShell 5.1 | Python 3.10.11 | 1 | `(3, 10)` |

`404 passed, 2 skipped` on `test_powershell_artifacts.py` + `test_supervisor.py`.

### Separately found, NOT fixed here

The required native-argument-quoting sweep turned up a **pre-existing** defect that is not PS5.1-specific — generated `bin/agenttalk.cmd` forwards raw `%*`, so `--task-arguments` and `--completion-json` values containing quotes or spaces are corrupted under **both** shells (JSON quotes dropped; one path value split into three argv entries). Deliberately not folded into a one-line quoting fix — it is a transport decision and is tracked as its own task.

### Process note

This guard was recommended by its reviewer as a **fast-follow**; I overrode that and had it built into this PR. The override is what put a defect in an already-approved diff. The guard now carries an executed dual-shell regression, and a third reviewer — neither the finder nor the builder — is reviewing the fix.

zoolok17:
## Round 4: the Python 3.11 guard is removed — `ce86fa4` → `c7ce6f3`

The connector surfaced **four fresh findings** on `ce86fa4`, and two of them exist *because of* the 3.11 preflight guard:

- **`supervisor.py:6528`** — the version probe lives only in `Preflight`, but the `ephemeral_launch` workflow calls `Launch-Spec` **directly**. So with `tool_runtime_python` on 3.10, an approved ephemeral reviewer still launched with `PYTHONSAFEPATH=1` that 3.10 ignores, and a package in the reviewer's working directory could shadow the pinned install — the exact hazard the guard exists to prevent, still open on a path that ordinary launches now reject.
- **`supervisor.py:5522`** — the tutorial documents 3.11+ as *conditional*, which is correct without the guard and false with it. The guard made its own surrounding docs wrong.

That is **partial-channel closure**: one launch path gated, another left open. Same defect shape this PR chain has spent the week closing.

### Why it comes out rather than getting patched

This guard was recommended by its reviewer as a **fast-follow**. I overrode that and folded it in here. It has since produced three defects: a total break of the feature under Windows PowerShell 5.1 (embedded double quotes stripped by 5.1's native-argument passing, so a genuine Python 3.14 was rejected), a documentation contradiction, and an unclosed launch path. Three defects from one late addition is the answer to whether the reviewer's sequencing judgment was right.

It now has its own task, where the channel gets enumerated properly — `Preflight`, `Launch-Spec`/ephemeral, `Launch`, the `.cmd` shim, `_child_env` — with a version-boundary test **per path** and execution under both shells.

### The removal is verifiable, not surgical

`supervisor.py`, `tests/test_supervisor.py` and `tests/test_powershell_artifacts.py` are restored **byte-for-byte** to `987a2d0` — the SHA independently approved before the guard was added:

```
$ git diff --stat 987a2d0 c7ce6f3
 src/agenttalk/python_import_env.py | 3 +++
 1 file changed, 3 insertions(+)
```

`python_import_env.py` is deliberately **not** reverted: its delta is the `PYTHONPYCACHEPREFIX` exclusion rationale answering a reviewer residual, unrelated to the guard.

So this PR is now the twice-approved channel-class work plus one docstring. The 3.10/`PYTHONSAFEPATH` residual returns to being an honestly documented limitation — the state that was reviewed and approved.

### The other two fresh findings are filed, not folded

- **`wrapper/run.py:597`** — `doctor` resolves `agenttalk_py` from `launch.windows_file` and prepends the checkout `PYTHONPATH`, so it probes a different executable and environment than the pinned launch. Both directions are live (OK on a broken pin; warn when the launch would work). This divergence *is* introduced here — before the pin, doctor and launch agreed by construction. It is a diagnostic defect rather than a hole in the pin, since `Preflight` still smoke-tests the real invocation and fails closed, which is why it is a follow-up.
- **`supervisor.py:7103`** — a non-ASCII runtime path embedded in a BOM-free UTF-8 `.cmd` is decoded by `cmd.exe` with its active code page, breaking every shim-based control command. Grouped with the pre-existing raw-`%*` forwarding defect in the same shim, because both are one question: what is the shim's argument and encoding contract, and should it exist for these calls at all.

Neither is being forced into a PR that is finally in its cleanest reviewed state — that is the mistake this round exists to undo.

PR #97, opened 2026-07-30
fix(supervisor): surface startup kill-switch holds

zoolok17:
## Summary

Fixes #114 by replacing the generated supervisor's silent startup `exit 3` with a checked, observational kill-switch record before the process exits.

- The observational path does **not** claim the executor instance or acquire an executor token. It authenticates generated-supervisor ancestry, holds the existing lifecycle → PowerShell-selection → config lock order through one field-owned mutation, re-checks the raw `supervisor.kill` file inside that transaction, and atomically records an idempotent activation observation.
- Startup and mid-poll activation use the same operation, so both activation orders have one persistence contract.
- Human `agenttalk status` and `status --json` project the raw hold and structured observation independently of the event-ring limit. A switch-present supervisor is now distinguishable from a supervisor that never started.
- `exit 3` is intentionally preserved after the startup observation because Scheduled Task retry treats it as a blocked failure.
- No process/config mutation is performed in observational startup mode, and `dependencies = []` remains unchanged.

## Root cause

The generated PowerShell scaffold tested `.agenttalk/supervisor.kill` before instance claim, config load, planning, or any persistence path. When the switch already existed at process start, the supervisor vanished with code 3 and left no operator-readable evidence.

A first implementation added a checked observation path but accepted only direct PowerShell and one `cmd.exe` hop. Windows wheel CI exposed that as a false-DOWN: a CPython venv can add its `Scripts\python.exe` redirector before the running base Python, and an installed console entry point can add the exact active `agenttalk.exe`. The shipped wheel path was therefore legitimate but refused.

The checked path now accepts only eight bounded, ordered launcher topologies built from the exact system `cmd.exe`, exact active `agenttalk.exe`, and exact active venv redirector: direct; each single hop; the three valid two-hop combinations; and `cmd → console → venv`. Every intermediate image is bound by native file identity, every process must remain active, creation times must be ordered, cycles and excess depth are rejected, and unattributed/reordered/copied launchers still fail closed.

The wheel regressions also remove checkout `PYTHONPATH` whenever pytest imported an installed candidate and assert the subprocess resolves `agenttalk.supervisor` to the same candidate. This prevents a packaged-code test from passing by silently importing checkout source.

## Operator impact and rollout

This changes the generated scaffold. Existing installations must refresh it with:

```text
agenttalk supervise --refresh-scripts
```

Safe refresh sequence: stop and wait for the current supervisor, remember whether the kill switch is armed, temporarily remove the switch if necessary so the refresh mutation is permitted, run `--refresh-scripts`, restore the switch before restart when it was previously armed, then start the supervisor.

This PR did not touch the live `.agenttalk/supervisor.ps1`, live supervisor, gateway, or ports 4000/4001.

## Validation

Candidate: `fc8e6818419c96e679bf5275146d8937fbbb3473` (clean before/after, stable HEAD/tree).

Focused and review evidence:

- Combined source supervisor/lifecycle/portability sweep: **419 passed, 2 skipped**.
- Authority/portability suites on Python 3.10 and 3.14: **43 passed** per runtime.
- Installed-wheel provenance + generated-shim regressions: **5 passed** on Python 3.10 and **5 passed** on Python 3.14, with imports resolving inside each temporary wheel venv's `site-packages`.
- All eight admitted ancestry tuples have positive coverage; copied images, identity mismatch/replacement, arbitrary Python, unattributed console, reordered/duplicate/excess-depth chains, cycles, start inversion, and exited launchers have negative coverage.
- Independent authority and test reviews: **approved with no blocking findings**.
- Ruff and `git diff --check`: pass.

Exact-SHA local release dev-gate, run `e23ca9277234467aaf6b5edb7de38073`:

| Required pytest leg | Status | Result | Duration |
|---|---:|---:|---:|
| source / Python 3.10 | pass | 4,018 passed, 22 skipped | 1,268.07s |
| wheel / Python 3.10 | pass | 4,005 passed, 35 skipped | 1,176.83s |
| source / Python 3.14 | pass | 4,018 passed, 22 skipped | 1,269.76s |
| wheel / Python 3.14 | pass | 4,005 passed, 35 skipped | 1,083.26s |

The normalized local verdict is `block` only because `gitleaks` is unavailable (`blocked_dependency`): **18/19 required checks passed**. Ruff, Bandit, pip-audit, Semgrep, Zizmor, package build, both wheel installs, dependency checks, wheel contracts, and initial/final SHA bindings all passed. Evidence: `D:\tmp\at114-gate-fc8e681-r2.json`; SHA-256 `be43850ad55ef782b220ca950c486f5a89d43e5c5215e1979eed83dedf1a8720`.

The unchanged 2,700-second pytest cap had substantial headroom: all four full legs completed in 18:03–21:09. That is direct evidence that the earlier 41%-at-cap observation was host contention, not an intrinsically undersized cap.

Authoritative GitHub Actions for this SHA:

- tests run `30564620345`: **success**, including Windows 3.10, 3.11, 3.12, and 3.13 plus the full declared Linux/macOS matrix and aggregate.
- security run `30564620246`: **success**.

## Known limits

- A kill switch removed and recreated entirely between two observations is indistinguishable from one continuous activation; no filesystem observer can recover an interval it never sampled.
- If an active raw switch coincides with a malformed or future-version observation sidecar, the raw hold still remains visible and automation stays blocked, but the structured record requires a switch cycle, manual repair, or newer compatible code before it can be rewritten.
- Alternate third-party Windows launchers outside the positively identified CPython venv/distlib topologies intentionally fail closed until separately characterized and tested.


zoolok17:
Closing in favour of **#107**, which supersedes this PR by its own description and by content.

#107 contains both of this branch's commits (`fix(supervisor): surface startup kill-switch holds`, `fix(supervisor): validate Windows launcher ancestry`) plus five more, and those five exist specifically to remediate the two connector P2s this PR still carries **new on its own head** at 15/15 green:

- the startup claim race — the CLI precheck is unlocked and the lifecycle/config-locked claim never re-reads `supervisor.kill`, so a switch appearing inside the precheck-to-claim window lets the claim succeed and keep executor authority;
- `read_runtime_observation` raising `OverflowError` instead of a typed invalid-record error on a syntactically valid but oversized epoch.

Green CI with unaddressed head-anchored findings is exactly the state our merge gate exists to refuse, so this branch was never mergeable as it stood — it only looked mergeable.

Keeping both open costs review attention twice on the same change and risks the older, less-correct one being merged because it is the one that is green. #107 needs a rebase (currently conflicting); that is tracked and dispatched.

No work is lost — every commit here is an ancestor of #107.

PR #98, opened 2026-07-30
Preserve bounded wrapper diagnostics across supervisor relaunches

zoolok17:
Fixes #117.

## Why

The generated supervisor launches wrappers in hidden windows without redirecting stdout or stderr. An unhandled wrapper exception therefore leaves no traceback, and a naive redirect would overwrite the dead wrapper's evidence on relaunch.

## What changed

- Redirect stdout and stderr to distinct files only for exact supervisor-launched `agenttalk wrap` commands.
- Allocate an immutable log generation before launch, then retain that generation plus the three newest prior committed generations.
- Fail soft: filesystem or cleanup failures warn and continue launching with a unique fallback generation or, if necessary, without redirection.
- Bound Python-level output to 1 MiB per stream across four fixed segments.
- On Windows, launch through an explicit handle allowlist so supervisor pipes and locks are not inherited.
- Emit timestamped factual lifecycle events such as `turn_started`, `child_spawned`, `child_exited`, `turn_ended`, and wrapper termination. These records contain no self-assessment and are not consumed by supervisor health or restart decisions.
- Keep `dependencies = []` unchanged.

This branch is based on current `master` and does not depend on #114 / PR #97.

## Log location and #108

This repository ignores `.agenttalk/`, but `agenttalk init` does not provision that ignore rule in adopter repositories. Logs therefore live outside the checkout in a per-user, project-hashed directory:

- Windows: `%LOCALAPPDATA%\agenttalk\wrapper-logs\...`
- POSIX: `$XDG_STATE_HOME/agenttalk/wrapper-logs/...` or `~/.local/state/agenttalk/wrapper-logs/...`

The logs can contain model output, tool output, and tracebacks and should be treated as sensitive.

## Operator action

This changes the generated PowerShell scaffold. After upgrading, regenerate it with:

```text
agenttalk supervise --refresh-scripts
```

Do not copy the source template over an existing scaffold manually.

## Validation

- `tests/test_supervisor.py`: 374 passed, 2 skipped
- wrapper log/runtime/profile suite: 95 passed, 1 skipped
- wrapper/lead-loop regression slice: 331 passed
- CLI wrap slice: 2 passed
- affected gate regression slice: 10 passed, 366 deselected
- targeted Ruff, `py_compile`, `git diff --check`, and Bandit: clean
- adversarial code and launch-matrix reviews: no P0/P1/P2 findings

Final SHA-bound dev gate for `9a1accc`:

- `complete=true`, `18/19` required checks passed; the normalized verdict is
  `block` only because `gitleaks` is unavailable (`blocked_dependency`).
- Python 3.10: source `4015 passed, 23 skipped`; wheel
  `4004 passed, 34 skipped`.
- Python 3.14: source `4015 passed, 23 skipped`; wheel
  `4004 passed, 34 skipped`.
- Ruff, Bandit, pip-audit, Semgrep, Zizmor, package build/install/dependency
  checks, wheel contracts, and initial/final SHA bindings all passed.
- Evidence: `D:\tmp\at117-gate-9a1accc.json`.

The four 4,038-test legs completed in 19:28-21:24, leaving substantial
headroom under the fixed 2,700-second cap. Together with the prior run, this
supports host contention—not an inherently undersized cap—as the explanation
for #114's earlier 41%-at-timeout observation; an idle-host rerun remains the
right confirmation.

## Known limits

- The byte bound begins at the Python entry point; interpreter/bootstrap output and direct native file-descriptor writes can bypass the cooperative stream wrapper.
- Fail-soft pending generations or unresolved filesystem errors can temporarily exceed the retention target.
- OOM, hard kill, or power loss cannot emit a terminal event; the gap after the last factual lifecycle record is the available evidence.


zoolok17:
@codex review

zoolok17:
@codex review

zoolok17:
@codex review

zoolok17:
@codex review

zoolok17:
**Disposition for the `supervisor.py:10535` finding (bound retention of unresolved log generations): DEFERRED, tracked as task #139, deliberately — not an oversight.**

This is the known and disclosed consequence of the prediction-to-observation change in this PR, not a new defect. Recording the reasoning so a future reader does not have to reconstruct it:

- **Before this PR**, a launch that never reached `cmd_wrap` (a configured `wrap --help`, say) **committed an empty generation and pruned real diagnostics.** That destroyed evidence.
- **After this PR**, the same launch leaves an *uncommitted* generation, which the retention rule preserves and which never evicts proven evidence. That leaks disk.

Destroying evidence is strictly worse than leaking disk, and the leak is visible as accumulating directories rather than silent. So the change is a strict improvement even with this residual open.

Closing the residual needs the launched PID recorded alongside the `.pending` marker (it is not recorded at creation time today — the PID is not known until `Start-WrapperProcess` returns) plus a prune rule that discards a pending generation once its owning PID is confirmed dead without ever having committed, reusing the project's existing PID-liveness tracking with a grace period wide enough not to prune a slow-but-legitimate start. That is medium-sized, touches retention logic, and is scoped on **#139** rather than folded here.

The honest residual, stated plainly: a wrapper that starts, authenticates, then dies before its first turn leaves a pending generation that is preserved and unpruned until #139 lands.

zoolok17:
@codex review

zoolok17:
**Disposition for the two round-27 findings at `c7b9d9f9`, both addressed in `14b20da`:**

1. *Document the extra retained log generation* — **you were right and our report was wrong.** Settled by execution, not by re-reading: quota=4, ten real `New-WrapperLogTargets`/confirm cycles against the real template, counting directories on disk. Counts were `[1,2,3,4,5,5,5,5,5,5]` — it reaches quota+1 the moment it first exceeds quota and never returns to quota. That is steady state, not the one-cycle transient we claimed. Root cause is structural: `Invoke-WrapperLogRetentionPrune` runs before the generation it just created is confirmed, so trim always precedes add and the add always lands one over. Reserving a slot does not fix it — it reintroduces evicting confirmed evidence for an unconfirmed guess, at any quota. The real fix must move the prune trigger to after a generation's fate is known, which is a mechanism change we are not folding into this PR unreviewed. Tracked as its own task. What `14b20da` does: rewrites `docs/supervisor-tutorial.md:161-177`, which was doubly stale (it also still described the pre-`ae3e749` commit-on-PID mechanism), to state quota+1 as a known accepted violation rather than a promise, plus a regression test pinning the exact trajectory so the claim cannot drift back to unverified.

2. *Preflight prefix flags with the configured invocation* — **fixed.** Preflight's wrapped-mode probe now splices in the declared prefix tokens via the same `Resolve-AgenttalkModuleFlagIndex` resolution every other `module_args_from` consumer uses, gated on `launch_mode -eq 'wrap'`. The gating is load-bearing: resolving unconditionally made a legacy-direct agent (whose args never invoke `-m agenttalk`) fail closed as a config error and refuse to launch — caught by the full suite, not by the new tests. Your finding also falsified the standing `-E`/`-I`/`-S` allowlist justification, so we re-examined it rather than only repairing the probe: the flags stay allowed, but the justification is now written as an explicit *dependency on Preflight fidelity* rather than a property the allowlist may assume on its own.

One boundary we are flagging rather than quietly clearing: `Launch-Spec` (ephemeral launches) has no Preflight call at all, for this or any other misconfiguration class. Pre-existing, broader than this PR, tracked separately.

@codex review

zoolok17:
**Disposition for round 30 at `d9e0db3`.**

**1. `dev-gate (linux/3.12)` was a real semgrep finding against code added in this PR, and it is a false positive — verified, not asserted.**

Rule `agenttalk-no-raw-agent-name-in-filename` fired on `f"{self.base_path.name}.cursor"` at `src/agenttalk/wrapper_logs.py:277`. The rule is a purely syntactic pattern matching `f"...{$NAME}.cursor"` / `.heartbeat`, written to catch an *unvalidated agent name* reaching a path (the v0.2.1 path-traversal class). It matches on shape regardless of what `$NAME` holds. Here `$NAME` is this class's own fixed `base_path`, always literally `stdout.log` or `stderr.log`, set once from `ENV_STDOUT_PATH`/`ENV_STDERR_PATH` and never agent-name-derived; the agent-name-to-path validation the rule protects already happened one layer up, where the supervisor hashes the agent name into the directory component. Note also that `Path.name` yields only the final component, so it cannot carry a directory traversal even if the source path did.

Confirmed by running semgrep locally against the real rule file: exactly 1 result at that line without a suppression, 0 with. The suppression added is **scoped** — `# nosemgrep: agenttalk-no-raw-agent-name-in-filename` — deliberately not a bare `# nosemgrep`, so nothing else on that line becomes silently exempt later.

**2. The `7a4ee43` cursor-persistence finding was still OPEN at `42542f9` and is closed at `d9e0db3`.** The reviewer was right that the append-only change did not cover the advance path: a stale cursor names a full `.1`, the resumed instance appends correctly, immediately hits capacity, and `_advance_tail` rotates into `.2` — and that rotation still used `wb` unconditionally. Reproduced with a direct test before any fix (hand-built full `.1`, real never-recorded `.2`, cursor stuck at `0`; old code truncated `.2`).

**3. It is closed by a general invariant rather than a patch to that path**, because this is the fifth distinct mechanism on this PR that reached the same outcome — losing the newest bytes. The earlier four: an unbounded second tee, a size-assumption trade-off, mtime ambiguity in resume, and the missing-cursor fallback.

The invariant: **`_open_tail` may open a suffix with `wb` only if this instance has itself already opened that exact index before, in its own memory, this run.** One attribute checked and updated at the single place segments get opened; no per-path special-casing. Truncation is no longer a function of index *selection* at all — only of this instance's own visit history — which is why stale, missing and correct cursors all become safe by the same rule, and why the rotation cascade lands safely on `.2` by construction rather than by a case added for cascades. It also allowed deleting the one-shot resume-append flag and reducing `_resume_tail_position` to "pick a starting index".

Stated limit, which is the ring working rather than a hole: after a full lap, a second visit to an index does truncate — discarding content this same instance already superseded.

@codex review

PR #102, opened 2026-07-30
fix(#120): register the owned process tree per agent (bounded, role-tagged, nonce-anchored)

zoolok17:
Implements task #120. Authored by codex-agenttalk-developer-3; pushed and opened by the lead because the agent's environment policy refused the egress destination (it correctly refused to bypass the control and escalated instead).

**Gated at clean SHA 1d3819dd86ccf9da598fa262770208a5d9fb3e8d — dev-gate reported 19/19 PASS by the author. Authoritative CI on this PR is still required before merge.**

## Why

The durable channel already existed (\managed_pids\ is persisted per supervisor poll). The real gap was narrower and sharper: **ownership traversal deliberately stopped at shell hosts**, so \wrapper -> codex -> pwsh -> node\ lost the pwsh/node branch — and the tool descendant is precisely what wedges a wrapper. When an agent hung on 2026-07-30 the lead had to enumerate the tree by hand from \Win32_Process\ to discover the codex child was already gone.

This evolves the existing per-agent projection into a bounded \owned_process_tree\. It does **not** add a schema, a writer, or a second kill path.

## Design

- **Ownership requires triple agreement** — supervisor state, the strict wrapper-runtime record, and the live wrapper — over \(pid, start)\, wrapper generation, and the launch nonce re-read from the live command line.
- **Names and command lines may LABEL an already-owned row; they may never establish ownership or kill authority.** Command-line-only and name-only authority is *removed* from the wrapped destructive path rather than unioned with the tree.
- Deterministic roles: \wrapper\, \cli_launcher\, \cli_brain\, \	ool_descendant\, with parent edges and \discovered_at\.
- **Bounded at 64 rows**, anchors always retained, deterministic deepest-descendant retention with ancestor closure, and cap+1 recording exact omission.
- **A truncated or invalid tree is HOLD evidence and authorises no automatic teardown** — a partial kill could strand omitted descendants, and a stranded descendant is exactly what wedges a wrapper.
- Kill integration reuses the existing leaves-first \Stop-Tree\, whose recycled-PID skip (live start-time mismatch => never kill) remains mandatory. The turn-watchdog kill path is untouched.

## Test coverage

Shell-crossing \codex -> pwsh -> node\; role/order/refresh without progress; every anchor mismatch; PID reuse; name/CommandLine spoof refusal; cap and cap+1 deterministic truncation and HOLD; current-snapshot start rebinding; and the real \Stop-Tree\ leaves-first start-guarded sink.

## Scope

Two commits, based on \2fc175\. Verified **not** to contain the unmerged #115 state-lock work (\ce5eba9\ is not an ancestor). Independent of the #114/#115 chain.

## Known open item

A tree exceeding the 64-row cap currently authorises no teardown, which would make a runaway spawner permanently unkillable by automation — the same fade-to-silence shape as #78. The lead required that a truncated tree **escalate** to a durable operator-visible item rather than merely HOLD; confirm whether that landed in this revision.

Unblocks #78 and #116.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -56,13 +56,15 @@ design. The item that did ship is the one measured to cost the most in the field
   line; process names never authorize teardown. Windows launcher handoff uses a
   nullable, generation-fenced `GetProcessTimes` lifetime certificate; when that
   certificate is present, an exited launcher's first child must carry an exact
-  FILETIME strictly inside the launcher's creation/exit interval. Tree-entry
-  `start_filetime` is also nullable; null supplies no exact FILETIME proof, so
-  PID/start remains required, and a missing current FILETIME is ambiguous when
-  the prior identity recorded one. A prior complete tree can bridge an exited
-  intermediate process only for the same generation and nonce, with the exact
-  previously recorded child identity and parent edge; a new or reparented child
-  fails closed. A complete tree feeds the existing leaves-first `Stop-Tree`.
+  FILETIME strictly inside the launcher's creation/exit interval. Authoritative
+  `complete`/`absent` Windows tree entries require an exact `start_filetime`;
+  `invalid`/`truncated` HOLD entries may retain null as readable failure
+  evidence, but null grants no identity authority. Linux boot-ID/start-ticks
+  tokens are exact without FILETIME. A missing current FILETIME is ambiguous
+  when the prior identity recorded one. A prior complete tree can bridge an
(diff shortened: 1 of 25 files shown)


zoolok17:
@codex review

zoolok17:
@codex review

zoolok17:
@codex review

zoolok17:
@codex review

zoolok17:
@codex review

PR #108, opened 2026-08-02
fix(#150): a confirmed-absent wrapper must be launchable under a sticky invalid owned-process-tree

zoolok17:
## Root cause, in one sentence

A one-second glitch permanently disabled fleet-wide auto-recovery: a single transient bad row in one poll's whole-host process snapshot (most likely a process exiting mid-enumeration so CIM never resolved a parent for it) failed the snapshot integrity check for that ONE poll, marking every currently-polled agent's `owned_process_tree` invalid at once - and that invalid status was then STICKY forever, because nothing ever re-checked it against a later, clean snapshot. The transient trigger is almost incidental; the defect is that an invalid record could never be re-derived, so a wrapper later confirmed completely absent still couldn't reach a relaunch decision. This PR fixes the stickiness, not the trigger.

## Summary (round 1)
- Reproduces the production incident end to end: one UNRELATED row anywhere in a poll's whole-host snapshot (e.g. a system process with a null/negative parent_pid) fails `_snapshot_pid_integrity_error` for that whole poll and marks EVERY currently-polled agent's `owned_process_tree` invalid that cycle - even though each agent's own row is clean.
- That invalid status used to be permanently sticky: `_unverified_owned_process_tree` never re-derived absence for a non-`{complete,absent}` prior, so a wrapper later confirmed **completely absent** from an empty snapshot still never reached a relaunch decision.
- Verified this closes BOTH gates that read the persisted status (`_plan_one`'s own short-circuit and `evaluate_launch_barrier`'s post-kill barrier), without touching Stop-Tree (#146 owns that path).

## Round 2: connector found a blocker in round 1's own eligibility guard

- **Blocker**: `omitted_count == 0` (round 1's eligibility bar) only ever counted the gap between admitted nodes and size-cap truncation - never a candidate branch the walk SAW and explicitly excluded for another reason (an exited CLI launcher with no exact lifetime that left a live child behind). Reproduced and confirmed the exploit against the actual round-1 code before fixing: this promoted straight to `absent` even with a live, orphaned child still unaccounted for - the duplicate-wrapper hazard, arriving through round 1's own fix.
- Closed by adding one new persisted field, `rejected_count` (schema bumped 2->3), incremented at every site in `_owned_process_tree`'s walk that excludes a specific candidate rather than admitting it - enumerated and classified site by site. Eligibility is now `omitted_count == 0 AND rejected_count == 0 AND at least one entry`.
- Round 1's own remedy message had two more problems: it told operators to confirm only the NAMED processes were gone (missing what `--acknowledge-owned-processes-stopped` already covers for truncated records), and it named `--reset-process-tree-ownership` for a case (an entryless placeholder) where that command's own generation/nonce check will always refuse it. Both corrected - the entryless case now says plainly that no scripted remedy is verified to work, rather than naming one that doesn't.
- `_current_proof_failed`'s `owned_process_tree_refreshed` flag had a stale hand-copied duplicate of the eligibility check that never got updated when round 1 widened it - ephemeral reviewers held (never archived) on a promoted-invalid tree. Fixed by removing the duplicate and trusting the callee's own return value.

## Round 3: the round-2 schema bump was itself a defect, plus one more class closed by construction

- **Blocker, retraction of round 2's own fix**: bumping `schema_version` 2->3 to force re-validation of legacy records reintroduced, fleet-wide at upgrade, the exact sticky HOLD this PR exists to remove - a v2 record's field-set mismatch alone (not the version number) made `_valid_owned_process_tree` reject it outright, so even a healthy wrapper's fresh walk got seeded with `prior_record_invalid` on the first poll after upgrade and stuck from the second poll onward. Reverted `schema_version` to 2. `rejected_count` is now an OPTIONAL field: a v2-shaped record (missing it) is still valid, readable authority - missing means UNKNOWN, never coerced to zero, so it stays ineligible for absence-promotion without needing to be invalid. This IS the migration; there's nothing left for a version bump to gate.
- The validator parsed `rejected_count` but never constrained it against `status` - added the missing constraint (complete/absent with a nonzero `rejected_count` now fails validation).
- Three more sites lost a candidate without incrementing `rejected_count` (a duplicate pid silently picked in `_snap_index`; a Windows row missing exact `start_filetime` admitted by the walk but separately refused by the validator; a rebuild after an untrusted-but-readable prior unable to reach an orphaned intermediate's live child). Per the explicit review instruction, replaced hand-enumeration with a mechanical invariant instead of patching a fourth site: **discovered must equal admitted plus rejected**, checked once at the end of the walk. `discovery_roots` (wrapper, launcher, brain, every prior entry regardless of trust) expand through the CURRENT snapshot's raw parent-child edges; any descendant absent from `owned` becomes a rejection.
- The reset remedy said "every process this record could name" - wrong in the same direction as round 2's own finding 2. A rejected candidate is exactly what this record's entries never named either, so the reset command's own identity list can't cover it. Added an explicit clause naming this when `rejected_count > 0`.
- Two existing tests failed mid-implementation and were fixed by excluding `recycled_parent_pids` roots from the discovery-closure expansion: a confirmed-recycled (replacement) process's current children are someone else's tree, not an unaccounted loss of ours.

## Round 4: two self-inflicted-by-round-3 gaps, plus a second producer of invalid records

- The `_snap_index` dedup (round 3's own finding-1 fix) popped an excluded pid's row out of `idx` before `_children_map(idx)` was built, destroying the only edge connecting it to its own parent - so neither the main walk nor round 3's discovery-closure invariant could ever reach it, or anything below it. Added a permissive parent->children map built straight from the raw snapshot, used only by the discovery-closure block to seed/expand its frontier; `idx`/`_children_map(idx)` (admission, identity lookups) are unchanged.
- Round 3's recycled-root exclusion in the discovery-closure was too blanket - it treated every current child of a recycled pid as foreign, when `evaluate_launch_barrier` already draws an old-side/new-side line for exactly this situation (a child whose exact start strictly predates the replacement belongs to the process this agent actually owned). Applied that same comparison instead of excluding wholesale; the two existing tests that forced the round-3 exclusion still pass unchanged.
- The one that mattered: the post-kill launch barrier and ephemeral teardown barrier (the `.ps1` template) are a SECOND producer of invalid `owned_process_tree` records - they mutate a formerly-complete tree's status/reason_code directly in PowerShell, entirely outside `_owned_process_tree`'s own walk, and never ran any accounting of their own. They left `rejected_count` at its stale "complete" value (0), so `process_tree_invalid_post_kill_*` sailed through the eligibility bar even though the barrier's own survivors were never counted. Fixed by removing the property entirely at both mutation sites - the same "missing means unknown, not zero" shape the v2-migration fix already established, reused for a field this record's producer never computed. No Python-side change needed; the existing missing-key handling already treats this correctly.
- Did not plumb the barrier's survivor observations through as real rejections (an open option, not ordered this round) - flagged as a real schema/CLI extension to `evaluate_launch_barrier`'s output, not required to close this blocker.

## Round 5: one more malformed-row gap closed, one closed only partially with the sizing question raised rather than answered speculatively

- A PRIOR entry's own current row can have a malformed `parent_pid` this poll (a transient WMI/CIM hiccup). `_snap_all_edges` correctly cannot build an edge for it, but "cannot place" was silently collapsing to "not ours": a discovery ROOT is deliberately never itself checked against `owned` (only its descendants are, since a root simply exiting is normal), so a childless malformed-parent root sat at `rejected_count 0` forever unless something else happened to name it. Fixed the bounded case: every prior entry's current row is now checked for `parent_pid` validity unconditionally, independent of whether `prior_authority` ended up trusted (the existing trusted-rehydration block's own `prior_parent_drift` check already covered the trusted case).
- Did **not** implement the fully general "every valid-PID row on the host is a candidate" reframing as literally proposed - a row's placement is entirely determined by its own `parent_pid` field, and when that's malformed with no other valid edge connecting it to a root (never a prior entry, never the declared launcher), there is no snapshot-only signal distinguishing "malformed row near our tree" from "malformed row elsewhere on a busy host." Treating every such row as ours would reintroduce round 1's original bug in a new, eligibility-blocking form for any host where such a row persists across many polls. Flagged the tension and asked for direction rather than building the general version speculatively.
- The round-3 remedy clause only fired for a KNOWN positive `rejected_count`. A legacy v2 record (missing the field entirely, deliberately still valid authority per the v2-migration fix) got NO warning at all - the operator was told to verify only the named/omitted identities when the walk that produced the record never counted rejections in the first place. Added a second branch: `rejected_count is None` now warns that whether anything was excluded is UNKNOWN, not zero, and must be treated the same as a nonzero count.

## Round 6: the residual from round 5, closed - it turned out to be bounded after all

- Round 5's fix only inspected `parent_pid` on a prior entry's row when `idx.get(pid)` actually returned one. When that pid's row is DUPLICATED this poll, `_snap_index` correctly excludes it entirely (ambiguous, not absent) - `idx.get(pid)` then returns `None`, indistinguishable from a genuinely exited descendant, and round 5's check silently skipped past it. The connector's own framing settled the bounding question raised in the round-5 reply: this is a PREVIOUSLY OWNED identity - a prior entry - so there is no ambiguity about whether it's ours; we recorded it. That signal (which the fully-general reframing declined in round 5 was missing) does not exist for an unrelated row elsewhere on the host, so this fix cannot reintroduce round 1's incident.
- Split `_snap_index` into a thin wrapper over a new `_snap_index_and_excluded`, which also returns the set of pids it excluded - distinct from a pid simply never present (absent, normal, already safe). `_snap_index`'s own signature and behavior are unchanged for its other 8 call sites. `_owned_process_tree` now rejects a prior entry whose current row is `None` AND in the excluded set; one that's simply absent (not excluded) is unchanged.
- First test attempt reused round 4's own duplicate-pid shape (parent_pid pointing at the launcher) - round 4's discovery-closure already catches that via the permissive edge map regardless of the pid's own exclusion, so the test passed on the PRE-FIX commit too. Caught it before committing (same discipline as the round-4 recycled-root false-green) and rebuilt with an unrelated parent_pid so only `prior_entries` seeding it as a root can reach it.

## Round 7: the named finding, plus a full survey of every other `_snap_index` consumer

- **The finding**: a snapshot containing two rows for an ACTIVE EPHEMERAL REVIEWER's launcher pid (even identical rows) made `_snap_index` exclude it, and `_ephemeral_process_alive` read the resulting `idx.get(pid)`-is-`None` the same as "process exited" - permanently misclassifying a healthy review as failed, since a persisted `held_terminal` is deliberately unrewritable by any later report. Fixed: `_ephemeral_process_alive` now returns `None` (unknown) rather than `False` for an excluded pid; the caller already only treats `alive is False` as a terminal-failure signal. Test proves the operator-visible outcome (the review survives the glitched poll, `held_terminal` stays unset), not the internal flag value.
- This was a regression from round 6's own `_snap_index`/`_snap_index_and_excluded` split: it taught exactly one caller (the prior-entry check) the new distinction, and the thin `_snap_index(snapshot)` wrapper let the other seven call sites keep reading a miss as a single fact. Surveyed all seven, each with a keep-or-fix reason recorded at its own call site:
  - Proven NOT to matter, by construction: the promotion check in `_unverified_owned_process_tree` and `evaluate_launch_barrier` are each gated by `_snapshot_pid_integrity_error` finding the same underlying condition first (a duplicated pid is a strict subset of what that whole-snapshot check already catches) - the excluded set is provably empty by the time either site's own fix would run. Found this by testing, not just reading: a first attempt at each site's fix passed identically with and without the change, exposing dead code before it was kept.
  - Classified `_attribution`'s launcher confirmation as genuinely indifferent by design - **round 8 found this wrong; see below.**
  - Fixed defensively, reachability not proven either way: `_unverified_owned_process_may_exist`, the wrapper-liveness check in `_wrapped_liveness`, and brain liveness in `_liveness` for the legacy, non-wrapped path. Every end-to-end scenario tried already reached the same safe outcome through a different, pre-existing mechanism first, so no test could discriminate these three branches' own effect - disclosed as such rather than shipped with a non-discriminating test standing in for real coverage.
- With every consumer surveyed and fixed, justified, or disclosed as unproven, the thin `_snap_index` wrapper has zero remaining callers and is retired - a future caller has no way to get an index without also being handed the one set this conflation kept breaking on.

## Round 8: the round-7 `_attribution` classification was wrong - classified by local behavior, not by what downstream consumes

- Round 7 read `_attribution`'s own code correctly (an excluded launcher pid only suppresses launcher confirmation) but generalized that reading to the whole function. Two more consumers of the SAME index a few lines below - both iterating `valid_priors`, the legacy attribution model's persisted `managed_pids` - apply the identical `idx.get(pid)`-is-`None` miss to a PRIOR-TRACKED DESCENDANT instead of the launcher: the kill-targeting loop drops the descendant's kill target when its row is merely duplicated (not absent) this poll, and the provenance loop drops it from `managed_by_pid` the same way, so the NEXT poll's `valid_priors` no longer carries it at all even once the ambiguity clears.
- The observable consequence: a manual or stale-heartbeat recovery kills the launcher (the only thing still targeted), the duplicate glitch clears before the post-kill snapshot, and the untargeted descendant is left alive alongside a freshly launched replacement - the duplicate-wrapper hazard again, on the legacy non-wrapped path.
- Fixed both, same rule as everywhere else: an excluded prior entry is now targeted for killing using the prior record's own pid/start (the only identity this attribution model carries for it) via a synthetic row - Stop-Tree's own start-time guard decides whether to actually kill it; and an excluded prior entry is kept unchanged in `managed_by_pid` rather than dropped, preserving it as ambiguous HOLD evidence across the glitched poll.
- Did not extend this to the BFS descendant-walk a few lines below (also reads `idx.get(parent_pid)` the same way) - not named by the connector, and closing it correctly needs a substitute for `_strict_child_edge`'s own start-time-ordering check when the parent row itself is absent, a larger, separate change flagged rather than built speculatively.
- Re-examined the two round-7-reverted dead-code sites and the three fixed-on-principle sites under the sharper question ("what does everything downstream consume", not "what does this site do"): the reverted pair's own output is either a pass-through of an already-correct prior tree or consumed once within the same poll cycle, never persisted or re-derived from later - the argument holds downstream too. No new discriminating scenario found for the other three within this round.

## Round 9: a double-count, a third remedy that refuses the operator, and one genuinely unbounded residual

- **The double-count**: round 3's own "harmless, `rejected_count` is a gate not a tally" justification for `mark_rejected` was true when written and stopped being true the moment round 3 (and round 5) added an operator-facing message that reports the number as a count - a value's tolerance for imprecision is a property of its CONSUMERS, not the value itself. A candidate rejected once during the main walk (e.g. an unproven virtual-parent descendant below a trusted-but-exited intermediate) is also absent from `owned`, so the discovery-closure loop found it again and called `mark_rejected` a second time for the identical pid - one lost candidate reported as two. `mark_rejected` now takes an optional pid and dedupes against a `rejected_pids` set; every call site with an identifiable single candidate now passes it. Verified the exact shape named: a trusted-but-exited intermediate with a brand-new, never-recorded live child now reports `rejected_count == 1`, and the message says "excluded 1 candidate(s)" - asserted the message text, not the counter.
- **The third remedy that refuses the operator**: a configured-agent `--reset-process-tree-ownership` (cli.py) requires the CURRENT wrapper runtime record to read `STATUS_VALID` before it will act, independent of the tree's own entries/`rejected_count`. Rounds 2 and 3 each closed the ONE precondition that had actually bitten (an entryless placeholder's missing nonce; a rejected candidate outside the identity list) by patching that specific case - which is why there was a third: a tree can hold real entries inherited from an earlier complete poll while the runtime record is exactly what made THIS poll invalid. One predicate now gates naming the command as an actionable recommendation: this poll's own liveness read of the same runtime record the command re-reads at reset time. When it's not valid, the message still names the command for context but never recommends running it, and points at operator investigation instead.
- **The malformed-pid residual**: a row whose own pid field is malformed (not `parent_pid` this time) contributes no edge either direction. Traced both sub-cases by execution: for a PRIOR-TRACKED intermediate this is already safe (the trusted-rehydration block reconstructs its identity from the prior record's own start/filetime, bypassing the malformed current row entirely - confirmed by execution, locked in with a regression test). For a first-time-seen row, this is genuinely unbounded - every other fix in this family works by looking a row up BY its own intact pid field, and that is exactly the field missing here; there is no snapshot-only signal distinguishing it from an unrelated malformed row elsewhere on the host. Documented as a residual in `_snap_all_edges`'s own comment with a test that demonstrates the gap by execution, per the connector's own offer to accept this as documented rather than have a fourth algorithm guessed at.

## Round 10: the round-8 deferred BFS gap, now concrete; and a fourth refusing remedy caused by an imprecise predicate

- **The deferred BFS gap, built**: round 8 fixed `_attribution`'s kill-targeting and provenance loops for an excluded prior entry (a tracked pid whose row is merely duplicated, not gone) but explicitly deferred the BFS descendant-walk a few lines below, which seeds from the SAME excluded pid yet still looked it up via `idx.get(parent_pid)` - `None` for the identical reason - and silently skipped it as a BFS parent. The connector's concrete example: a tracked pid with duplicate rows (excluded) has an opaque, never-recorded child underneath it; round 8's fix makes the tracked pid a kill target again, but the child is never even queried, let alone targeted, because the skip happens before `kids.get(parent_pid, [])` is ever consulted. Consequence: the tracked pid gets killed, the opaque child survives it (the launch barrier has no tree evidence to block on for this legacy path either), and a replacement launches alongside a live child of the pid it just replaced - the exact hazard this PR exists to prevent.
- Resolved the deferred design question rather than deferring it again: `_strict_child_edge` only ever reads `parent["pid"]` and `_start_of(parent)` (`parent["start_time"]`) - the same two fields the excluded seed's synthetic row (round 8's own construction) already carries and that Stop-Tree already trusts to order that pid's own kill. The synthetic row is therefore sufficient to stand in as the BFS parent too; built a small `seed_synthetic_rows` map alongside `seed_pids` in the kill-targeting loop and consulted it in the BFS loop only when `idx.get(parent_pid)` misses - every other pid that ever reaches the queue was admitted via a real `idx` row already, so this fallback is scoped exactly to the excluded-seed case, not a general substitute for a real row.
- Verified the operator-visible outcome the lead named, not a changed internal set: a tracked pid with duplicate rows and an opaque child underneath it now has BOTH killed in the same operation - direction-controlled against 070d8dd.
- **The fourth refusing remedy**: round 9's fix gated the reset remedy on `liveness.get("runtime_status") == STATUS_VALID` - the lead's own "one predicate" instruction, but a single SIGNAL rather than the command's actual admission predicate. `--reset-process-tree-ownership` (`process_tree_ownership_reset_evidence`) separately requires the runtime record's own `wrapper_pid`/`wrapper_start` to agree with the stored `launcher_pid`/`launcher_start` - a record can be `STATUS_VALID` (it parses) while still failing that agreement, which is exactly what sends `_wrapped_liveness` down `wrapper_state_mismatch` with a nonempty inherited tree. Extracted the actual shared conjunct into one function, `_wrapper_identity_matches_state(record, st) -> bool`, and call it from all three places that used to check this ad hoc: `_wrapped_liveness`'s own HOLD branch, `process_tree_ownership_reset_evidence`'s admission check, and the remedy-message predicate in `_plan_one` (reading `liveness["runtime_record"]`, the exact record `_wrapped_liveness` already normalized this poll - not a re-derivation). The `elif has_entries` remedy wording now distinguishes the two distinct refusal reasons (runtime not valid vs. valid-but-mismatched identity) instead of collapsing them into one message that would misreport a valid status as "not valid".
- Verified the operator-visible outcome: the recommendation string is absent when a `STATUS_VALID` record's wrapper identity mismatches supervisor state, reaching `process_tree_invalid_wrapper_state_mismatch` with real entries - direction-controlled against 070d8dd.
- Caught and corrected a false-confirming-green during this round's own work: the first cut of the BFS test ran against the MAIN checkout's installed `supervisor.py` (editable install), not this worktree's file, because `git stash`-ing the worktree source has no effect on what gets imported without `PYTHONPATH` pointed at the worktree - the test "passed" identically with the fix present and stashed away. Caught by checking `sup.__file__` directly rather than trusting the stash to isolate anything; every direction-control in this round re-ran with `PYTHONPATH` set to the worktree's `src`.

## Round 11: a zero that survived a walk that never happened; one more excluded-vs-absent consumer; and a sizing decision on the closing question

- **The stale zero**: `_current_proof_failed` converts a previously COMPLETE tree to `invalid` (because THIS poll's runtime record is invalid or wrapper-mismatched, not because of anything about the tree's own entries) via `_unverified_owned_process_tree`'s "held" fallthrough - and that conversion is a relabeling, not a walk. It used to carry the original walk's `rejected_count == 0` forward unchanged, so a LATER poll's `entries_are_complete` eligibility check read that stale zero as proof THIS invalid record was itself complete, re-admitting it for absence-promotion on an accounting question nobody asked this poll. Same rule round 4 already applied to the PowerShell post-kill barrier's own mutation sites ("a producer that doesn't walk must not claim to have walked") - now applied to this Python-side analogue. Fixed by popping `rejected_count` from the converted record; missing reads as unknown, which is already ineligible for re-derivation - the correct meaning, since this poll genuinely does not know. Surveyed the other `copy.deepcopy(prior)`-shaped sites in this file for the same shape: the function's own top-of-function passthrough (prior already ineligible, unchanged either way) and `_owned_process_tree`'s two truncated/invalid passthroughs return prior UNCHANGED (no status relabeling happens there, so nothing to strip); the entryless placeholder builder hardcodes `rejected_count: 0` but pairs it with `entries: []`, and `bool([])` is already falsy, so it can never satisfy `entries_are_complete` regardless of the count - confirmed harmless by construction, not merely by inspection. Left the sibling "absent" promotion (same function, a few lines above) unchanged: `eligible_for_rederivation` never consults `rejected_count` once status is already `absent`, no message path reads it for that status, and the validator accepts either `0` or missing there - proven inert, not merely assumed so.
- **One more excluded-vs-absent consumer**: `_children_map(idx)` is built only from `idx.items()` - a newly discovered (never a prior entry) child excluded this poll (duplicate/ambiguous rows) never contributed a row to `idx` at all, so it never contributes an edge either, regardless of whether its parent is a trusted, non-excluded seed. Structurally different from round 10's finding: an excluded PARENT still has an independently-trusted prior identity to fall back on (the prior record); a first-time-seen EXCLUDED CHILD has no such anchor - the ambiguous current-poll rows are the only information that exists about it, the same "no snapshot-only signal to disambiguate a first-time-seen identity" residual already documented in `_snap_all_edges` for a malformed pid. Inventing a synthetic child row from one of the ambiguous rows would extend trust exactly where round 1's original incident began, so this is not treated the same way as round 8/10's synthetic-row fix. Instead: the BFS loop now also walks the permissive `_snap_all_edges` map per visited parent, and any child pid found there that `idx` excluded gets bumped into a new diagnostic counter (`excluded_live_descendant_unaccounted`, deduped per pid) rather than silently vanishing - never invented as a kill target, never lost either.
- **The closing question**: asked whether the lookup itself could be restructured so "excluded" is impossible to ignore - a discriminated result a caller must destructure, rather than a set a caller can decline to consult. Judgment call, with reasons: not pursuing it this round. `idx`-shaped data has 42 total accessor touch points across this file (`.get`/`.pop`/`.items`/`.values`/subscript), of which only ~8 have ever needed the excluded-vs-absent distinction to make a correct decision - the rest are either indifferent by design or already paired with an explicit `excluded_pids` check. A discriminated-result API would need every one of those 42 sites migrated (including the ones already correct) to stay coherent, and even then Python has no compiler-enforced exhaustiveness check on match/branch handling - it would make the excluded case visible at each site, but the RIGHT handling (synthetic row vs. diagnostic-only vs. genuinely indifferent - three different, correct answers found across rounds 7-11) still requires the same per-site domain judgment a type change cannot automate. That is a large, invasive refactor for a partial benefit. Recommending: accept the current state as enumerated-and-fixed with a bounded, judged residual, not structurally impossible - continuing this file's own established convention of a proof-bearing comment at every indifferent/justified site is the cheaper mechanism doing most of the real work already.
- Both fixes verified against the operator-visible outcome, `PYTHONPATH` pinned throughout: the trusted root is still targeted while the ambiguous child is neither invented as a kill target nor silently lost (diagnostics counter == 1); and the stale-zero conversion no longer carries `rejected_count` forward. Direction-controlled against c05a2b1.

## Round 12: the diagnostic that was never read, a hazard our own round-11 fix created, and a proposed inversion of the whole shape

- **Fix, mandatory regardless of anything else**: round 11's `excluded_live_descendant_unaccounted` diagnostic was computed and never consulted - the legacy manual-restart path still killed only the trusted, accounted-for root and relaunched, exactly the outcome the diagnostic exists to prevent. A check that runs and is never acted on is not a check. Added a HOLD at the same precedence the wrapped path's `tree_status` check already has over even an authorized restart marker: when this diagnostic is nonzero, the legacy path now returns `WARN_ONLY`/`UNACCOUNTED_LIVE_DESCENDANT` (no kill, no relaunch, not even the trusted parent) instead of proceeding on partial accounting.
- **A hazard round 11's own fix created**: round 10's synthetic-row fallback (an excluded seed's own prior pid/start, reused as the `_strict_child_edge` PARENT for BFS descendant discovery) is safe as a KILL TARGET - Stop-Tree re-verifies the live process's actual start time against it before acting, protecting against a recycled pid. It is NOT safe as a time-ordering PARENT for discovering FURTHER descendants: nothing re-verifies that claim before it feeds an independent kill decision about a different pid, and a recycled pid's stale prior start time is almost always earlier than any of the replacement process's own children - the ordering check meant to reject an unrelated descendant would instead pass every one of them. Reverted round 10's synthetic-row reuse in the BFS loop; an excluded parent's children (real row or themselves excluded) are now treated the same way round 11 already treats an excluded first-seen child - unaccounted, never a kill target. This converges rounds 10, 11, and 12's fixes into one uniform rule: an excluded pid touching the discovery graph, as parent or as child, always produces an accounted-for diagnostic, never a kill target built on an unverified time-ordering claim - and the new HOLD above is what actually makes "unaccounted" safe rather than merely visible.
- **A third finding, confirmed but not fixed this round**: a malformed prior with a live wrapper rebuilds only the currently-reachable subset (the prior's own historical entries are discarded entirely, since they never parsed) as `prior_record_invalid`, with `rejected_count == 0` earned honestly by THIS narrowed walk - but that walk never had the chance to check on whatever else the FULL historical tree might have tracked before the malformation. The narrowed record reads as complete when it is only complete relative to a smaller graph than the one that matters. Confirmed by tracing the code, not fixed independently this round: fixing every instance of "a record claims completeness it didn't earn" one producer at a time is exactly the pattern that hasn't converged since round 9, and this is another instance of it, not a new site.
- **The proposed inversion**: replace "eligibility requires the absence of recorded rejections" (`rejected_count == 0`, which is the default state of an integer and therefore free to every new or narrowed code path) with "eligibility requires positive evidence a complete walk happened" - one boolean, `walk_complete`, computed ONCE at the true end of `_owned_process_tree`'s own walk (as `bool(entries) and omitted_count == 0 and rejected_count == 0 and not prior_record_invalid`, i.e. the exact same conjunction `entries_are_complete` already computes today, but stamped at the source instead of re-derived by every later reader from fields that can be inherited, defaulted, or hardcoded without ever having been jointly checked), never set anywhere else, and required present-and-true by `_unverified_owned_process_tree`'s eligibility check in place of the current `entries_are_complete` re-derivation.
  - Sized before recommending: `rejected_count`/`omitted_count` stay as diagnostics (accuracy still matters for the operator-facing message), but stop being independently load-bearing for the safety decision - the walk-complete boolean already folds them in at the moment they're freshly known, so a later reader never re-derives (and cannot mis-derive) the same fact from possibly-stale fields.
  - Answering the posed question directly: yes, there is a path that promotes to absent without a NEW walk running on that specific poll - the canonical case this whole PR exists to fix, a wrapper confirmed absent from an available current snapshot, checked directly against a PRIOR record's entries. This is not a carve-out that complicates the design: `walk_complete` is a persistent fact about the STORED record, earned on whichever earlier poll's walk actually produced it, not something that must be re-earned every poll it's read. The mechanism already works this way today (`entries_are_complete` is also read from `prior`, not recomputed fresh each poll) - the inversion changes WHERE the fact is computed (once, at the source) and HOW it's represented (an explicit claim instead of an inferred one), not WHEN it's allowed to be trusted.
  - This directly explains finding 3 (`3000`) without a separate fix: the narrowed `prior_record_invalid` rebuild is exactly the case where `entries_are_complete`'s existing conjunction reads true when it should not - `walk_complete` closes it by construction, since `not prior_record_invalid` is already one of its own terms.
  - Every OTHER current and future non-walking producer (placeholders, relabelings, the PowerShell post-kill barrier) is safe by omission under the inversion, not by remembering to strip a field - EXCEPT sites that `copy.deepcopy` an existing record forward without a NEW walk (this file's own `_unverified_owned_process_tree` "held" conversion, and the PowerShell barrier's status mutation) still need to explicitly strip `walk_complete`, the same care already applied to `rejected_count` at both those exact sites.
  - Sizing: touches `_owned_process_tree`'s result construction, `_unverified_owned_process_tree`'s eligibility check, the schema/validator (`walk_complete` optional, same v2-migration-safe pattern as `rejected_count`), and the PowerShell template's existing `rejected_count`-stripping mutation sites (would gain a sibling strip). Not built this round - sized and reported per the standing instruction to size before building; recommending it for the next round rather than folding a cross-cutting schema change into an already-large one.
- Both mandatory fixes verified against the operator-visible outcome, `PYTHONPATH` pinned throughout: the poll HOLDS (no kill, no relaunch, empty kill_targets) whenever a live descendant cannot be verified, whether the ambiguity is on the parent or the child side of the discovery graph. Direction-controlled against c74f776.

## Round 13: the safety HOLD we just built expires while the hazard it guards is still observable

- A new class for this PR: not "excluded conflated with absent" but "a TTL conflating unresolved-with-decayed, arriving through the time axis." An excluded prior seed that stays ambiguous (duplicate/conflicting rows) longer than `_PROVENANCE_TTL_SECONDS` (one hour) was preserved unchanged by round 8's fix, but its own `last_fresh_attribution_epoch` was never touched - `_prior_valid`'s TTL check eventually expires it, even though the pid is present in conflicting rows on every single poll in between. For a seed no longer connected to the launcher, dropping it also removes the only root that discovers its child, so round 12's `UNACCOUNTED_LIVE_DESCENDANT` HOLD disappears and a pending authorized restart relaunches without targeting either process.
- A TTL is the right instrument for evidence nobody has rechecked; it is the wrong instrument for an unresolved condition still being observed every poll. Fixed by refreshing `last_fresh_attribution_epoch` on every poll the ambiguity is still directly observed (the excluded branch of the `managed_pids` persistence loop), so the entry cannot lapse while the condition producing it is still present.
- A first attempt generalized the refresh to the validly-matching (non-excluded) branch too, on the reasoning that both branches are equally never re-derived via `_new_provenance_entry` (`prior_only` is set unconditionally for both). This broke an existing, deliberately-pinned test proving a valid match's TTL is meant to age normally - it has other paths to re-earn a fresh stamp (`own_wrapper`/`own_wait`/`live_chain_descendant`/`legacy_rederived`) that an excluded identity does not. Caught by running the full suite before reporting the round done, not assumed correct from the reasoning alone; reverted the valid-branch change and documented explicitly why it stays excluded-only.
- Surveyed the file for other TTL-style checks that might really be unresolved-condition markers: `_PROVENANCE_TTL_SECONDS` is the only safety-gating TTL found. The launch-coordination barrier's own `last_observed_epoch` already refreshes on continued observation of the same blocking condition - the correct pattern, already present elsewhere, not something this round needed to invent. Runtime heartbeat/progress-age staleness is legitimately decaying evidence (an unrefreshed heartbeat really does mean nobody's heard from the process), not an unresolved-condition marker - a different, correctly-TTL'd shape.
- No interaction with the proposed `walk_complete` inversion: `_prior_valid`'s TTL belongs exclusively to `_attribution` (the legacy, non-wrapped path); `walk_complete` would belong exclusively to `_owned_process_tree`/`_unverified_owned_process_tree` (the wrapped path). `_liveness` dispatches each agent to exactly one of the two, never both - disjoint data, disjoint mechanisms, confirmed by tracing the dispatch rather than asserted from the file's own shape.

## Test plan
- [x] Round 1: unit tests on `_unverified_owned_process_tree` (invalid-but-complete promotes; truncated/no-entries never promote; still-alive stays held) + end-to-end `plan_actions`/`evaluate_launch_barrier` test.
- [x] Round 2: the blocker scenario, built and confirmed to actually promote to absent under the round-1 code before being fixed; refreshed-flag, entryless-message, and truncated-remedy tests, each direction-controlled against 43a0484.
- [x] Round 3: schema-migration end-to-end (a v2-shaped prior must not brick a healthy wrapper); validator status/rejected_count constraint; `_snap_index` duplicate-pid exclusion; admit/reject consistency for missing exact filetime; discovery-closure orphan reconciliation; reset-remedy wording - each direction-controlled against 89858e0.
- [x] Round 4: `_snap_index` exclusion no longer orphans a duplicated pid's descendants from the discovery-closure invariant; recycled-root discovery agrees with `evaluate_launch_barrier`'s old-side/new-side line instead of excluding every current child (the two pre-existing recycled-root tests still pass unchanged); post-kill/ephemeral-teardown barrier mutation no longer presents a trustworthy zero `rejected_count` - each direction-controlled against 059a9b7.
- [x] Round 5: a prior entry's malformed-parent-pid current row is now rejected rather than silently unaccounted; the legacy-v2 remedy warns on an unknown `rejected_count` the same as a known nonzero one - each direction-controlled against ca6c7f3.
- [x] Round 6: a prior entry whose current row `_snap_index` excluded (not merely absent) is now rejected too - direction-controlled against ac48478, with a corrected fixture after catching a false-confirming first attempt.
- [x] Round 7: an ephemeral reviewer with a duplicated launcher-pid row now survives the poll instead of being archived failed - direction-controlled against ca0f5a0. The other six surveyed sites are either proven dead code (verified by a discriminating test), genuinely indifferent by design, or fixed defensively with the reachability gap disclosed rather than papered over with a non-discriminating test.
- [x] Round 8: a legacy agent's duplicated-but-tracked descendant is now a kill target on a manual-restart poll (closing the hazard) and survives a routine poll's provenance (verified independently, since a restart poll legitimately resets `managed_pids` on its own) - direction-controlled against 3d8c2d6.
- [x] Round 9: a trusted-but-exited intermediate's brand-new live child now reports `rejected_count == 1` and "excluded 1 candidate(s)" in the message, not 2 (direction-controlled against d8c5000); the reset remedy no longer recommends a command that would refuse the operator when the current runtime record is invalid, even with real entries present (direction-controlled against d8c5000); the malformed-pid residual's two sub-cases (prior-tracked: safe; first-time-seen: unbounded) are each verified by execution and behave identically on d8c5000, confirming they characterize existing/documented behavior rather than a code change.
- [x] Round 10: an opaque child of an excluded-but-tracked pid is now a kill target in the same operation as that pid, closing the deferred BFS gap (direction-controlled against 070d8dd); the reset remedy no longer recommends the command when a `STATUS_VALID` runtime record's wrapper identity mismatches supervisor state (direction-controlled against 070d8dd) - both verified with `PYTHONPATH` pointed at this worktree's `src` after catching a first pass that silently ran against the main checkout's installed package instead.
- [x] Round 11: converting a complete tree to invalid no longer carries the original walk's `rejected_count` forward, closing a re-admission gap on a later poll (direction-controlled against c05a2b1); a newly discovered, excluded (duplicate-row) child of a trusted, non-excluded root is neither invented as a kill target nor silently lost - it surfaces as a new diagnostic counter instead (direction-controlled against c05a2b1).
- [x] Round 12: the legacy path now HOLDS (no kill, no relaunch) whenever `excluded_live_descendant_unaccounted` is nonzero, instead of relaunching on partial accounting; the round-10/11 BFS tests updated to assert this HOLD directly, and confirmed to fail against c74f776 (the old code still relaunches / still invents the child as a kill target via the stale-time hazard) - both direction-controlled against c74f776. Finding 3 (`prior_record_invalid` narrowed rebuild reading complete) confirmed but intentionally not fixed independently this round; sized alongside the proposed `walk_complete` inversion instead.
- [x] Round 13: an excluded prior seed that stays ambiguous longer than `_PROVENANCE_TTL_SECONDS` (one hour) now has its `last_fresh_attribution_epoch` refreshed on every poll the ambiguity is still directly observed, instead of expiring while the pid is present in conflicting rows on every single poll - a TTL is the right instrument for evidence nobody has rechecked, the wrong one for an unresolved condition still being actively observed. Scoped to the excluded branch only: a first attempt also refreshed a validly-matching (non-excluded) prior's epoch, which broke an existing, deliberately-pinned test proving a valid match's TTL is meant to age normally (it has other paths to re-earn a fresh stamp that an excluded identity does not) - caught by the full suite, reverted, documented. Surveyed the file for other TTL-style safety gates (none found beyond this one; the coordination barrier's own `last_observed_epoch` already refreshes correctly on continued observation) and confirmed no coupling with the proposed `walk_complete` inversion (disjoint mechanisms - `_prior_valid`'s TTL belongs to the legacy `_attribution` path, `walk_complete` would belong to the wrapped `_owned_process_tree` path, and a given agent is dispatched to exactly one). Direction-controlled against 32668ee.
- [x] Direction-controlled throughout: every new/changed assertion fails against the code it's fixing (verified by stashing `supervisor.py` alone and re-running).
- [x] Full `tests/test_supervisor.py` suite green on the default interpreter and `py -3.10` (544 passed, 2 skipped, 1 xfailed - 543 + 1 new).
- [x] `ruff check` clean on both touched files.
- [ ] dev-gate/build/wheel/security legs intentionally not run locally (targeted tests only, per standing directive) - left to CI.

### DIFF
--- a/src/agenttalk/supervisor.py
+++ b/src/agenttalk/supervisor.py
@@ -4754,6 +4754,18 @@ def _attribution(
             continue
         row = idx.get(entry["pid"])
         if isinstance(row, dict) and _start_of(row) == entry["start"]:
+            # Deliberately UNCHANGED (see
+            # test_process_ownership_provenanced_prior_exact_fields_
+            # request_and_ttl): a validly-matching prior is not itself
+            # ambiguous, so its own last_fresh_attribution_epoch is left
+            # to age normally toward _PROVENANCE_TTL_SECONDS - the TTL
+            # bounds how long an identity may be passively carried forward
+            # without a fresh, independently-earned re-derivation, and an
+            # unambiguous match is exactly the case that DOES have another
+            # path to re-earn one (own_wrapper/own_wait/live_chain_
+            # descendant/legacy_rederived). Round 13 only touches the
+            # EXCLUDED branch below, where no such alternative path exists
+            # while the ambiguity persists - see that branch's own comment.
             managed_by_pid[entry["pid"]] = entry
         elif entry["pid"] in excluded_pids:
             # Task #150 round 8 connector finding: idx.get(pid) is None
@@ -4762,9 +4774,33 @@ def _attribution(
             # managed_by_pid here is what actually persists the loss: the
             # NEXT poll's valid_priors would no longer carry this
             # identity at all, even once the ambiguity clears. Keep it
-            # unchanged as ambiguous HOLD evidence rather than silently
-            # forgetting a still-tracked descendant.
-            managed_by_pid[entry["pid"]] = entry
+            # as ambiguous HOLD evidence rather than silently forgetting a
+            # still-tracked descendant.
+            #
+            # Task #150 round 13 connector finding: keeping the entry
+            # UNCHANGED left its own last_fresh_attribution_epoch frozen at
+            # whenever it was last unambiguous - _prior_valid's TTL check
+            # then expires it after _PROVENANCE_TTL_SECONDS of continuous
+            # ambiguity, even though the pid is present in conflicting rows
+            # on EVERY one of those polls. Unlike the matching branch
+            # above, an excluded identity has NO alternative path to earn
+            # a fresh stamp while the ambiguity persists - own_wrapper/
+            # own_wait/live_chain_descendant/legacy_rederived all require
+            # a real idx row, which this pid does not have by construction.
+            # A TTL is the right instrument for evidence nobody has
+            # rechecked; it is the wrong instrument for an identity
+            # rechecked every poll that just cannot be disambiguated - a
+            # live observation of an unresolved condition, not aging
+            # evidence. Expiring it converts "we still cannot tell" into
+            # "there was never anything here", dropping the seed the very
+            # UNACCOUNTED_LIVE_DESCENDANT HOLD depends on to discover its
+            # own child. Refresh the epoch on every poll the ambiguity is
+            # still directly observed, so the entry cannot lapse while the
+            # condition producing it is still present.
+            managed_by_pid[entry["pid"]] = {
+                **entry,
+                "last_fresh_attribution_epoch": now_epoch,
+            }
 
     for legacy in legacy_priors:
         pid = legacy.get("pid")
--- a/src/agenttalk/supervisor.py
+++ b/src/agenttalk/supervisor.py
@@ -2073,6 +2073,36 @@ def _snap_index(snapshot: list[dict] | None) -> dict[int, dict]:
     return idx
 
 
+def _snap_all_edges(snapshot: list[dict] | None) -> dict[int, list[int]]:
+    """Permissive parent -> children edges, straight from every row's own
+    pid/parent_pid - including a pid _snap_index excluded for identity
+    ambiguity (a duplicate or malformed row).
+
+    Task #150 round 4 connector finding 1: _snap_index's exclusion popped
+    the excluded pid's row out of idx BEFORE _children_map(idx) was built,
+    which destroyed the only edge connecting it to its own parent - so
+    neither the main walk nor the discovery-closure invariant could ever
+    reach it, or reach anything further below it, to reject it. A
+    candidate the walk cannot trust to ADMIT is still a candidate it must
+    ACCOUNT for. This map is never used for admission or identity lookups
+    (idx/_children_map(idx) still own that, unchanged) - only to seed what
+    the discovery-closure invariant reconciles against, where routing
+    through an untrusted edge is exactly the point.
+    """
+    kids: dict[int, list[int]] = {}
+    for row in snapshot or []:
+        if not isinstance(row, dict):
+            continue
+        pid = row.get("pid")
+        parent_pid = row.get("parent_pid")
+        if not isinstance(pid, int) or isinstance(pid, bool):
+            continue
+        if not isinstance(parent_pid, int) or isinstance(parent_pid, bool):
+            continue
+        kids.setdefault(parent_pid, []).append(pid)
+    return kids
+
+
 def _snapshot_pid_integrity_error(snapshot: list[dict]) -> str | None:
     """Validate the PID/parent graph before any last-row-wins indexing.
 
@@ -3097,6 +3127,11 @@ def _owned_process_tree(
     add_node(wrapper_row, role="wrapper", depth=0, live=True)
     wrapper_pid = wrapper_row.get("pid")
     recycled_parent_pids: set[int] = set()
+    # Task #150 round 4 connector finding 2: the replacement's own row, kept
+    # alongside the pid, so the discovery-closure invariant can draw the
+    # same old-side/new-side line evaluate_launch_barrier already draws
+    # instead of excluding every child of a recycled root.
+    recycled_parent_rows: dict[int, dict] = {}
 
     # Rehydrate exact live identities from the previous complete tree. A live
     # descendant whose intermediate parent exited remains owned only because
@@ -3128,6 +3163,7 @@ def _owned_process_tree(
                     and _has_exact_start_identity(row)
                 ):
                     recycled_parent_pids.add(entry["pid"])
+                    recycled_parent_rows[entry["pid"]] = row
                     continue
                 if not isinstance(row, dict) or not _start_tokens_match(
                     _start_of(row),
@@ -3195,6 +3231,8 @@ def _owned_process_tree(
         )
         if current_launcher_identity == "different":
             recycled_parent_pids.add(launcher_pid)
+            if isinstance(current_launcher, dict):
+                recycled_parent_rows[launcher_pid] = current_launcher
         launcher_live = bool(
             isinstance(current_launcher, dict)
             and current_launcher.get("parent_pid") == wrapper_pid
@@ -3366,11 +3404,11 @@ def _owned_process_tree(
     # simply exited) of a known root - the wrapper, the declared launcher,
     # the discovered brain, and every PRIOR entry, whether or not
     # prior_authority ended up trusted - reachable via the CURRENT
-    # snapshot's raw parent-child edges, regardless of whether admission
-    # logic actually walked into it. children.get(pid, ...) still returns a
-    # live child whose OWN row names pid as parent even when pid itself is
-    # absent or was never visited, so an orphaned intermediate's live child
-    # - or a rejected launcher's live child - is still IN this closure. Any
+    # snapshot's raw parent-child edges (discovery_edges below), regardless
+    # of whether admission logic actually walked into it. A live child whose
+    # OWN row names pid as parent is found even when pid itself is absent
+    # or was never visited, so an orphaned intermediate's live child - or a
+    # rejected launcher's live child - is still IN this closure. Any
     # such descendant that never made it into "owned" is unaccounted -
     # either a case already flagged above (harmless double count;
     # rejected_count is a gate, not an exact tally) or exactly the silent-
@@ -3389,28 +3427,57 @@ def _owned_process_tree(
         _prior_pid = _prior_entry.get("pid") if isinstance(_prior_entry, dict) else None
         if isinstance(_prior_pid, int) and not isinstance(_prior_pid, bool):
             discovery_roots.add(_prior_pid)
-    # A root already PROVEN to be a different, recycled process (exact
-    # identity mismatch, not merely absent) is not a candidate this agent
-    # has any claim to - its current children are a REPLACEMENT process's
-    # own descendants, not ours, and the existing recycled_parent_pids
-    # check a few lines above already treats them as foreign on purpose
-    # (test: test_owned_process_tree_recycled_launcher_ignores_foreign_
-    # children / ..._recycled_virtual_parent_ignores_replacement_child).
-    # Expanding into a recycled root's children here would flag someone
-    # else's process tree as our own unaccounted loss.
+    # Task #150 round 4 connector finding 1: children.get(...) above is
+    # _children_map(idx) - built from the DEDUPED idx, so a pid _snap_index
+    # excluded for identity ambiguity (a duplicate or malformed row) has no
+    # row left in idx to contribute its own edge to that map. Its parent
+    # never learns it exists, and neither does anything below it - the
+    # exact silent loss this invariant exists to close, just moved one
+    # level earlier than round 3 checked. discovery_edges is built straight
+    # from the raw snapshot instead, so an excluded pid's edge to its OWN
+    # parent survives for discovery/rejection purposes (never for
+    # admission - idx/children, used everywhere above this block, are
+    # unchanged and still the strict, trusted source).
+    discovery_edges = _snap_all_edges(snapshot)
+    # Task #150 round 4 connector finding 2 (round 3's own exclusion was too
+    # blanket): a root already PROVEN to be a different, recycled process
+    # does not make EVERY current child foreign - evaluate_launch_barrier
+    # already draws the old-side/new-side line for exactly this situation
+    # (a child whose exact start strictly predates the replacement belongs
+    # to the process this agent actually owned; only a child at or after
+    # the replacement's own exact start belongs to the replacement).
+    # Missing/incomparable exact evidence stays conservative - treated as
+    # ours - matching the barrier's own default. Applying that same test
+    # here, rather than excluding wholesale, is what the two tests that
+    # forced the round-3 exclusion actually needed: they construct a
+    # replacement whose OWN current children postdate it, which this still
+    # excludes correctly.
     discovered_descendants: set[int] = set()
-    frontier = [
-        pid
-        for root in discovery_roots
-        if root not in recycled_parent_pids
-        for pid in children.get(root, [])
-    ]
+    frontier: list[int] = []
+    for root in discovery_roots:
+        if root not in recycled_parent_pids:
+            frontier.extend(discovery_edges.get(root, []))
+            continue
+        replacement_key = _exact_start_order_key(recycled_parent_rows.get(root))
+        for pid in discovery_edges.get(root, []):
+            child_key = _exact_start_order_key(idx.get(pid))
+            if (
+                replacement_key is not None
+                and child_key is not None
+                and child_key[0] == replacement_key[0]
+                and child_key[1] >= replacement_key[1]
+            ):
+                # At or after the replacement's own exact start: the
+                # replacement's own descendant, not a candidate we ever
+                # owned.
+                continue
+            frontier.append(pid)
     while frontier:
         candidate_pid = frontier.pop()
         if candidate_pid in discovered_descendants:
             continue
         discovered_descendants.add(candidate_pid)
-        frontier.extend(children.get(candidate_pid, []))
+        frontier.extend(discovery_edges.get(candidate_pid, []))
     for _ in sorted(discovered_descendants - set(owned)):
         mark_rejected("unaccounted_discovery")
 
@@ -11400,6 +11467,14 @@ $pollNum = 0
             $p.barrier_state.owned_process_tree.truncated = $false
             $p.barrier_state.owned_process_tree.omitted_count = 0
             $p.barrier_state.owned_process_tree.observed_count = [int]$p.barrier_state.owned_process_tree.recorded_count
+            # Task #150 round 4 connector finding 3: this barrier observed
+            # survivors by its OWN accounting, not the Python walk's - it
+            # must not leave a stale rejected_count (0, from whenever this
+            # tree was last complete) presenting as "the walk found nothing
+            # unaccounted". Removing the property entirely reads as unknown
+            # downstream, never zero - the same shape the v2-migration fix
+            # already relies on for a field a record simply never had.
+            $p.barrier_state.owned_process_tree.PSObject.Properties.Remove('rejected_count')
           }
           if ($p.barrier_state) { Set-AgentState $state $name $p.barrier_state } else { Set-AgentState $state $name $p.next_state }
           if (-not (Save-StateForPoll $state)) {
@@ -11616,6 +11691,10 @@ $pollNum = 0
             $p.next_entry.owned_process_tree.truncated = $false
             $p.next_entry.owned_process_tree.omitted_count = 0
             $p.next_entry.owned_process_tree.observed_count = [int]$p.next_entry.owned_process_tree.recorded_count
+            # Task #150 round 4 connector finding 3: see the matching
+            # comment on the relaunch/stuck_recover barrier above - same
+            # second-producer shape, same fix.
+            $p.next_entry.owned_process_tree.PSObject.Properties.Remove('rejected_count')
           }
           $p.next_entry | Add-Member -NotePropertyName process_tree_hold_reason `
             -NotePropertyValue ("process_tree_invalid_post_kill_{0}" -f $why) -Force
--- a/src/agenttalk/supervisor.py
+++ b/src/agenttalk/supervisor.py
@@ -2131,6 +2131,27 @@ def _snap_all_edges(snapshot: list[dict] | None) -> dict[int, list[int]]:
     the discovery-closure invariant reconciles against, where routing
     through an untrusted edge is exactly the point.
     """
+    # Task #150 round 9 connector finding (documented residual, not fixed):
+    # a row whose own PID field is malformed - as opposed to round 5's
+    # malformed PARENT_PID - contributes no edge either direction, since
+    # there is no valid pid to key it by. For a row bearing a PRIOR-ENTRY
+    # pid this is already safe: the trusted-rehydration block above never
+    # looks at this row at all - it reconstructs the intermediate's
+    # identity straight from the prior record's own start/filetime and
+    # re-verifies any live child against that same prior chain, bypassing
+    # the malformed current row entirely (confirmed by direct execution,
+    # not merely reasoned). For a row that has NEVER been a prior entry -
+    # no independently-trusted pid to reconcile it against - there is no
+    # snapshot-only signal distinguishing "this malformed-pid row is a
+    # lost candidate of ours" from "this malformed-pid row belongs to an
+    # unrelated process elsewhere on the host": every other fix in this
+    # family works by looking a row up BY its own intact pid field
+    # (idx.get(known_pid)), and that is exactly the field missing here.
+    # Treating every malformed-pid row as ours would revive the exact
+    # host-wide-poisoning failure mode round 5 already ruled out for the
+    # analogous malformed-parent_pid case. Genuinely unbounded for a
+    # first-time-seen intermediate; accepted as a residual rather than
+    # patched speculatively.
     kids: dict[int, list[int]] = {}
     for row in snapshot or []:
         if not isinstance(row, dict):
@@ -3104,9 +3125,29 @@ def _owned_process_tree(
     # exactly what makes the "invalid but re-derivable" case in
     # _unverified_owned_process_tree safe.
     rejected_count = 0
+    # Task #150 round 9 connector finding: rejected_count's own "harmless
+    # double count, this is a gate not a tally" justification (round 3)
+    # stopped being true the moment round 3 (and round 5) added an
+    # operator-facing message that reports the number - a value's
+    # tolerance for imprecision is a property of its CONSUMERS, not of
+    # the value, and a new, more precise consumer retroactively
+    # invalidates an old "close enough" judgement. A live child behind an
+    # unproven virtual-parent bridge (or any other candidate rejected
+    # once during the main walk) is also absent from `owned`, so the
+    # discovery-closure loop below finds it AGAIN and would double-count
+    # it. Dedupe by pid wherever a call site has one: the same candidate
+    # is only ever one exclusion, however many different checks noticed
+    # it. Calls with no single identifiable candidate (a whole-snapshot
+    # fact, or the mandatory-closure size cap) pass no pid and are never
+    # deduped against each other.
+    rejected_pids: set[int] = set()
 
-    def mark_rejected(reason: str) -> None:
+    def mark_rejected(reason: str, pid: int | None = None) -> None:
         nonlocal invalid_reason, rejected_count
+        if isinstance(pid, int) and not isinstance(pid, bool):
+            if pid in rejected_pids:
+                return
+            rejected_pids.add(pid)
         rejected_count += 1
         invalid_reason = invalid_reason or reason
 
@@ -3136,7 +3177,10 @@ def _owned_process_tree(
             or not start
             or start_key is None
         ):
-            mark_rejected("invalid_process_row")
+            mark_rejected(
+                "invalid_process_row",
+                pid if isinstance(pid, int) and not isinstance(pid, bool) else None,
+            )
             return False
         if not _has_exact_start_identity(row):
             # Windows' CIM timestamp is rounded. It can identify graph edges,
@@ -3156,7 +3200,7 @@ def _owned_process_tree(
             # both sides agree: an entry this imprecise never contributes to
             # a rejected_count of 0, so it can never reach a status the
             # validator would then refuse.
-            mark_rejected("exact_start_filetime_unavailable")
+            mark_rejected("exact_start_filetime_unavailable", pid)
         existing = owned.get(pid)
         if existing is not None:
             if (
@@ -3164,7 +3208,7 @@ def _owned_process_tree(
                 or _filetime_of(existing["row"]) != start_filetime
                 or existing["row"].get("parent_pid") != parent_pid
             ):
-                mark_rejected("pid_identity_collision")
+                mark_rejected("pid_identity_collision", pid)
                 return False
             existing["depth"] = min(int(existing["depth"]), depth)
             if role_rank[role] > role_rank[existing["role"]]:
@@ -3196,14 +3240,20 @@ def _owned_process_tree(
     # already earned. Missing ancestors are retained as non-killable bridges.
     if prior_authority is not None and prior_entries:
         prior_root = prior_entries[0]
+        prior_root_pid = prior_root.get("pid")
         if (
-            prior_root.get("pid") != wrapper_pid
+            prior_root_pid != wrapper_pid
             or not _start_tokens_match(
                 prior_root.get("start"),
                 _start_of(wrapper_row),
             )
         ):
-            mark_rejected("prior_wrapper_mismatch")
+            mark_rejected(
+                "prior_wrapper_mismatch",
+                prior_root_pid
+                if isinstance(prior_root_pid, int) and not isinstance(prior_root_pid, bool)
+                else None,
+            )
             prior_authority = None
         else:
             live_prior: set[int] = set()
@@ -3228,13 +3278,13 @@ def _owned_process_tree(
                 ):
                     continue
                 if row.get("parent_pid") != entry["parent_pid"]:
-                    mark_rejected("prior_parent_drift")
+                    mark_rejected("prior_parent_drift", entry["pid"])
                     continue
                 if not _filetime_identity_matches(
                     row,
                     entry.get("start_filetime"),
                 ):
-                    mark_rejected("prior_filetime_drift")
+                    mark_rejected("prior_filetime_drift", entry["pid"])
                     continue
                 live_prior.add(entry["pid"])
             needed = set(live_prior)
@@ -3320,7 +3370,7 @@ def _owned_process_tree(
                 )
                 and current_launcher.get("parent_pid") != wrapper_pid
             ):
-                mark_rejected("launcher_parent_mismatch")
+                mark_rejected("launcher_parent_mismatch", launcher_pid)
             launcher_row = {
                 "pid": launcher_pid,
                 "parent_pid": wrapper_pid,
@@ -3343,13 +3393,13 @@ def _owned_process_tree(
                     # authority.
                     launcher_row = None
                 elif current_launcher is not None:
-                    mark_rejected("exact_start_filetime_unavailable")
+                    mark_rejected("exact_start_filetime_unavailable", launcher_pid)
                     launcher_row = None
                 elif children.get(launcher_pid):
                     # Windows retains PPID edges after a parent exits. Without
                     # an exact lifetime/prior certificate those children are
                     # ambiguous.
-                    mark_rejected("unproven_virtual_parent_descendant")
+                    mark_rejected("unproven_virtual_parent_descendant", launcher_pid)
                     launcher_row = None
                 else:
                     # A childless, absent launcher cannot be killed and its
@@ -3357,11 +3407,11 @@ def _owned_process_tree(
                     # ownership authority.
                     launcher_row = None
         if launcher_pid == wrapper_pid:
-            mark_rejected("launcher_equals_wrapper")
+            mark_rejected("launcher_equals_wrapper", launcher_pid)
         elif launcher_row is not None:
             edge_reason = _strict_owned_child_edge(wrapper_row, launcher_row)
             if edge_reason is not None:
-                mark_rejected(edge_reason)
+                mark_rejected(edge_reason, launcher_pid)
             add_node(
                 launcher_row,
                 role="cli_launcher",
@@ -3380,7 +3430,7 @@ def _owned_process_tree(
         for child_pid in sorted(children.get(parent_pid, [])):
             child = idx.get(child_pid)
             if not isinstance(child, dict):
-                mark_rejected("missing_child_row")
+                mark_rejected("missing_child_row", child_pid)
                 continue
             if not parent["live"]:
                 prior_parent = prior_by_pid.get(parent_pid)
@@ -3423,11 +3473,11 @@ def _owned_process_tree(
                     # or by the exited launcher's exact lifetime bounds.
                     continue
                 if not certified_launcher_child and not prior_owned_child:
-                    mark_rejected("unproven_virtual_parent_descendant")
+                    mark_rejected("unproven_virtual_parent_descendant", child_pid)
                     continue
             edge_reason = _strict_owned_child_edge(parent_row, child)
             if edge_reason is not None:
-                mark_rejected(edge_reason)
+                mark_rejected(edge_reason, child_pid)
                 continue
             role = (
                 "cli_brain"
@@ -3446,7 +3496,10 @@ def _owned_process_tree(
     if isinstance(brain_row, dict):
         brain_pid = brain_row.get("pid")
         if brain_pid not in owned:
-            mark_rejected("brain_outside_owned_tree")
+            mark_rejected(
+                "brain_outside_owned_tree",
+                brain_pid if isinstance(brain_pid, int) and not isinstance(brain_pid, bool) else None,
+            )
         elif owned[brain_pid]["role"] == "tool_descendant":
             owned[brain_pid]["role"] = "cli_brain"
 
@@ -3467,10 +3520,15 @@ def _owned_process_tree(
     # or was never visited, so an orphaned intermediate's live child - or a
     # rejected launcher's live child - is still IN this closure. Any
     # such descendant that never made it into "owned" is unaccounted -
-    # either a case already flagged above (harmless double count;
-    # rejected_count is a gate, not an exact tally) or exactly the silent-
-    # loss class hand enumeration cannot bound. Either way it becomes a
-    # rejection here, mechanically, rather than by finding a fourth site.
+    # either a case already flagged above by pid (mark_rejected's own
+    # dedup means re-flagging the same pid here is a no-op, not a double
+    # count - round 9 connector finding retracted round 3's "harmless,
+    # rejected_count is a gate not a tally" call: that was true before
+    # round 3 itself added an operator-facing message reporting the
+    # number, and stopped being true the moment it did) or exactly the
+    # silent-loss class hand enumeration cannot bound. Either way it
+    # becomes a rejection here, mechanically, rather than by finding a
+    # fourth site.
     discovery_roots: set[int] = set()
     if isinstance(wrapper_pid, int) and not isinstance(wrapper_pid, bool):
         discovery_roots.add(wrapper_pid)
@@ -3495,8 +3553,9 @@ def _owned_process_tree(
     # a since-corrupted intermediate the prior once recorded would
     # otherwise sit at rejected_count 0 forever. This check is independent
     # of that trust gate on purpose - it does NOT retire the rehydration
-    # block's own check (a double-count there is harmless; rejected_count
-    # is a gate, not an exact tally).
+    # block's own check. Both pass the same entry pid to mark_rejected, so
+    # a poll where BOTH fire for the same identity is deduped by pid
+    # rather than double-counted (round 9 connector finding).
     for _prior_entry in prior_entries:
         _prior_pid = _prior_entry.get("pid") if isinstance(_prior_entry, dict) else None
         if not isinstance(_prior_pid, int) or isinstance(_prior_pid, bool):
@@ -3515,11 +3574,11 @@ def _owned_process_tree(
             # examines a pid that isn't already a recorded identity of
             # ours.
             if _prior_pid in excluded_pids:
-                mark_rejected("prior_entry_row_unindexable")
+                mark_rejected("prior_entry_row_unindexable", _prior_pid)
             continue
         _prior_parent_pid = _prior_row.get("parent_pid")
         if not isinstance(_prior_parent_pid, int) or isinstance(_prior_parent_pid, bool):
-            mark_rejected("prior_entry_parent_pid_malformed")
+            mark_rejected("prior_entry_parent_pid_malformed", _prior_pid)
     # Task #150 round 4 connector finding 1: children.get(...) above is
     # _children_map(idx) - built from the DEDUPED idx, so a pid _snap_index
     # excluded for identity ambiguity (a duplicate or malformed row) has no
@@ -3571,8 +3630,8 @@ def _owned_process_tree(
             continue
         discovered_descendants.add(candidate_pid)
         frontier.extend(discovery_edges.get(candidate_pid, []))
-    for _ in sorted(discovered_descendants - set(owned)):
-        mark_rejected("unaccounted_discovery")
+    for _unaccounted_pid in sorted(discovered_descendants - set(owned)):
+        mark_rejected("unaccounted_discovery", _unaccounted_pid)
 
     observed_count = len(owned)
     mandatory: set[int] = set()
@@ -7003,9 +7062,30 @@ def _plan_one(name: str, rpt: dict, st: dict, config: dict, cfg_agent: dict,
         # all. The flag's own text only mentions omitted identities; a
         # nonzero rejected_count needs the same warning stated explicitly,
         # not folded silently into "every process this record could name".
+        # Round 9 (connector finding, the third instance of the same wrong
+        # direction): the first two rounds each closed the ONE precondition
+        # that had actually bitten - an entryless placeholder's missing
+        # nonce (round 2), a rejected candidate the command's identity list
+        # can't see (round 3) - by patching that specific case, which is
+        # why there was a third: a configured-agent reset ALSO requires
+        # (cli.py, --reset-process-tree-ownership) the CURRENT wrapper
+        # runtime record to read STATUS_VALID, entirely independent of the
+        # tree's own entries/rejected_count. A tree can hold non-empty
+        # entries while the runtime record is exactly what made it invalid
+        # in the first place (process_tree_invalid_runtime_absent/invalid/
+        # record_missing/record_invalid) - the remedy named the command for
+        # precisely that case. One predicate now gates naming the command
+        # at all: this poll's own liveness read of the SAME runtime record
+        # the command re-reads at reset time. It can still go stale between
+        # this poll and when the operator acts - unavoidable from where the
+        # message is produced - but it is never wrong about what THIS poll
+        # already knows, which is what "must never guess" means here.
+        runtime_currently_valid = (
+            liveness.get("runtime_status") == runtime_obs.STATUS_VALID
+        )
         has_entries = bool(owned_tree.get("entries"))
         rejected_count = owned_tree.get("rejected_count")
-        if has_entries:
+        if has_entries and runtime_currently_valid:
             # Round 4 (connector finding 2, the same wrong direction a third
             # time): a legacy v2 record predates rejected_count entirely -
             # missing, not zero, is what keeps it out of absence-promotion
@@ -7044,6 +7124,18 @@ def _plan_one(name: str, rpt: dict, st: dict, config: dict, cfg_agent: dict,
                 "required --acknowledge-* flags); do NOT kill the wrapper "
                 "expecting an automatic relaunch - that is only safe once "
                 "this record already reflects the wrapper as absent"
+            )
+        elif has_entries:
+            remedy = (
+                "recovery: this record has named identities, but the "
+                "current wrapper runtime record is not valid "
+                f"({liveness.get('runtime_status')}) and --reset-process-"
+                "tree-ownership requires exactly that record to read valid "
+                "before it will act - it will refuse; this needs operator "
+                "investigation of "
+                f"{owned_tree.get('reason_code')} (a valid runtime record "
+                "recovering on its own is what would let a scripted reset "
+                "apply here, not a --acknowledge-* flag)"
             )
         else:
             remedy = (
--- a/src/agenttalk/supervisor.py
+++ b/src/agenttalk/supervisor.py
@@ -2746,14 +2746,37 @@ def _unverified_owned_process_tree(
 ) -> dict | None:
     """Fail closed when current proof cannot safely reuse a prior tree.
 
-    A complete prior remains the durable absence certificate when a current
-    snapshot proves every recorded pid/start identity absent. Any prior identity
-    still present, or an unavailable snapshot, converts it to sticky invalid
-    HOLD evidence. Existing truncated/invalid records are already sticky.
+    A complete or absent prior remains the durable absence certificate when a
+    current snapshot proves every recorded pid/start identity absent or
+    replaced. Task #150 connector finding: an INVALID prior is not
+    categorically excluded from that same re-derivation - only when its own
+    entries are a COMPLETE description of what was found (omitted_count == 0,
+    at least one entry recorded). Whatever made the tree invalid in the first
+    place (in the production incident, a single UNRELATED row elsewhere in
+    that poll's whole-host snapshot failing _snapshot_pid_integrity_error, not
+    anything about this agent's own identities) says nothing about whether
+    those SPECIFIC recorded identities are still alive - the same evidentiary
+    bar the complete/absent case already trusts. A TRUNCATED prior's entries
+    are incomplete by definition (omitted_count > 0) and never qualifies:
+    proving the recorded subset gone says nothing about whatever the size cap
+    dropped. A prior with no entries at all (schema-drift/generation-adoption
+    placeholders built by _invalid_owned_process_tree_record, never a real
+    tree walk) has nothing to disprove against and never qualifies either.
+    Any prior identity still present, or an unavailable/poisoned CURRENT
+    snapshot, remains sticky HOLD evidence exactly as before.
     """
     if prior is None:
         return None
-    if prior.get("status") not in {"complete", "absent"}:
+    entries = prior.get("entries")
+    entries_are_complete = (
+        isinstance(entries, list)
+        and bool(entries)
+        and prior.get("omitted_count") == 0
+    )
+    eligible_for_rederivation = prior.get("status") in {"complete", "absent"} or (
+        prior.get("status") == "invalid" and entries_are_complete
+    )
+    if not eligible_for_rederivation:
         held = copy.deepcopy(prior)
         if (
             held.get("status") == "invalid"
@@ -2776,10 +2799,12 @@ def _unverified_owned_process_tree(
                 expected_start=entry.get("start"),
                 expected_filetime=entry.get("start_filetime"),
             )
-            for entry in prior.get("entries", [])
+            for entry in entries or []
             if isinstance(entry, dict)
         ]
-        if all(state in {"absent", "different"} for state in identity_states):
+        if identity_states and all(
+            state in {"absent", "different"} for state in identity_states
+        ):
             absent = copy.deepcopy(prior)
             absent["status"] = "absent"
             absent["reason_code"] = "process_tree_absent"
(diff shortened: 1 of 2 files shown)


zoolok17:
Accepted, and your correction of my framing is right. Three times tonight I have named a mechanism from inference and been wrong; this is the third, and yours is the first account that actually reproduces.

## Your finding plus my measurement complete each other - and the synthesis matters

You proved the mechanism: ONE unrelated row anywhere in the whole-host snapshot fails `_snapshot_pid_integrity_error` for the entire poll, and that single result feeds every agent planned that cycle. Nine agents, one reason_code, one poll - as you say, that arithmetic only works whole-snapshot.

But note something that at first looks like a contradiction and is actually the last piece. When I sampled this host directly I found **zero** bad rows: 266 processes with no null CreationDate and no null ParentProcessId, and the live row-builder produced 276 rows with zero failing the predicate. So the poisoning row was NOT present when I looked. You used a synthetic `System` pid=4 row as a stand-in precisely because you could not point at a real one.

Put that together with the stickiness you found - an INVALID prior returns UNCHANGED regardless of what any later snapshot proves - and every observation fits one story:

- A **transient** bad row appeared in a single poll at 20:31:19Z. Probably a process exiting mid-enumeration so CIM could not resolve its parent. One poll, maybe one second.
- That one poll poisoned all nine agents at once, because the check is whole-snapshot.
- Every subsequent poll for the next several hours was clean - and it did not matter, because the invalid prior is never re-derived.
- Hence my sample finding nothing wrong, hence the identical reason_code across nine agents, hence dev-8 sitting absent through six poll decisions and never relaunching.

**So the transient cause is almost irrelevant and the stickiness is the entire defect.** A one-second glitch became a multi-hour fleet-wide loss of auto-recovery, and would have stayed that way indefinitely. That is exactly the seam you widened, so the fix is aimed at the right thing - please put this synthesis in the PR description, because "we relaxed a condition" reads very differently from "a one-second glitch permanently disabled the fleet and this is why."

If you can cheaply confirm the transient-row hypothesis from any retained snapshot or event detail around 20:31:19Z, do - but do not spend long on it. It is corroboration, not load-bearing: the stickiness is independently demonstrated and is what we are fixing.

## Your two pushbacks are both correct and I accept them

**P1** - you are right, `_plan_one` already persists a freshly-derived complete tree every poll; the write path exists. My "nothing writes the record" claim came from grepping persisted state for `owned_process_tree` and finding nothing, and I read absence-of-file as absence-of-code. What it actually meant was that this fleet had not reached "complete" since being poisoned, so there was nothing to write. That is the same error in a different costume - inferring a mechanism from a missing artifact.

**P3** - also right, and I am glad you did not guess a number. `notify` is rate-limited periodic, not permanently false; my complaint is really "this HOLD deserves a stronger cadence than a routine warn," which is a policy call. Leave it. I will decide the cadence separately rather than have it smuggled into a correctness fix.

The two you did take were the right two, and the reasoning on both is sound - `snapshot_` prefix matching the convention the file already uses for exactly this distinction, and naming `--reset-process-tree-ownership` in the message alongside an explicit warning against killing the wrapper. That warning is the single most valuable line in the change: it is the thing that would have stopped me making dev-8 worse.

## On the fix itself

The eligibility boundary is the part I would have got wrong. Promoting only an invalid prior whose entries are a COMPLETE description (`omitted_count == 0`, at least one entry), and refusing both a TRUNCATED prior and an entries-empty placeholder, is correct - and catching that `all(x for x in [])` is `True` would have silently re-opened a vacuous-truth hole in a naive relaxation. Verifying through the real `plan_actions` AND then separately through `evaluate_launch_barrier` is the right instinct too: proving the first gate opens is not proving the launch happens, and you checked the second gate rather than assuming they share a key.

Connector review requested on 43a0484. Stand by; nothing else queued for you.

zoolok17:
@codex review

zoolok17:
**Disposition for the four findings at `43a0484`, all addressed in `89858e0`.**

**Finding 1 (blocker) — confirmed real before being fixed, and closed as a class.** The hazard was reproduced against the pre-fix code first: a CLI launcher that has exited, carries no exact lifetime certificate, and left a live child behind produces `status=invalid, reason_code=unproven_virtual_parent_descendant, entries=[wrapper only], omitted_count=0`, and that genuinely promoted to `absent` against a later empty snapshot. So this was an exploit, not a missing field.

You correctly identified that `omitted_count` cannot carry the meaning the eligibility check needed — it counts admitted identities excluded by the cap, not rejected discovery branches. Rather than tighten three conditions, the fix adds one persisted field, `rejected_count`, fed by a single choke point (`mark_rejected`) called at every site in the walk that excludes a specific candidate. All sixteen `invalid_reason` assignment sites were enumerated by hand and classified; fourteen mark a rejection, two deliberately do not (a node admitted anyway with merely-imprecise identity, already covered by per-entry re-verification; and a childless absent launcher with nothing to strand). Ambiguous sites were counted as rejections on purpose: a false positive costs a longer hold, which is the existing safe default, while a false negative is the defect itself.

Eligibility is now `omitted_count == 0 AND rejected_count == 0 AND at least one entry`. `schema_version` is bumped 2→3 so records persisted before this change are treated as schema-drifted at upgrade — the existing cold-start path — rather than silently reading as if the new field were zero.

**Finding 4 — same root, one fix.** `_current_proof_failed` was re-testing `prior.status in {complete, absent}`, a hand-copied duplicate of the callee's own eligibility gate that went stale the moment that gate widened. The duplicate is removed rather than re-synchronised; `status == "absent"` after the call is already the complete signal. Because `_ephemeral_owned_process_view` goes through the same `_wrapped_liveness`, this also restores `teardown_ready` for ephemeral reviewers without a second patch.

**Findings 2 and 3 — the remedy was verified before being named.** `--reset-process-tree-ownership` requires the tree evidence's `wrapper_generation`/`launch_nonce` to agree with a current valid runtime record. For anything the walk actually produced (non-empty entries) those are the real current values and the command works; for an entryless placeholder they are usually null and the command's own check refuses it. The message now branches on that: non-empty entries get the reset, reworded so the precondition covers identities the record never named (matching `--acknowledge-owned-processes-stopped`'s stated contract rather than narrowing it); entryless gets an honest statement that no scripted remedy applies and why. **No remedy is named for the case where one could not be verified** — the previous message told operators to run a command that would refuse them.

Worth noting a test-construction correction made along the way: the first entryless scenario was wrong, because a healthy wrapper reaches the walk and gets real entries — the same `reason_code` text arising from a different mechanism. The entryless placeholder specifically requires a failed liveness check. That was caught by running it and reading the result, not by inferring the shape from the reason code.

All new tests direction-controlled against `43a0484`, including running the finding-1 scenario by hand there to confirm it promotes. `test_supervisor.py`: 521 passed, 2 skipped, 1 xfailed, both interpreters. Ruff clean.

@codex review

zoolok17:
**Disposition for the six findings at `89858e0`, all addressed in `059a9b7`.**

**Finding 2 — the schema bump was a defect, and the real mechanism was not the version number.** `_valid_owned_process_tree` rejects a v2 record through its strict `frozenset(value) == _OWNED_PROCESS_TREE_FIELDS` equality, regardless of what `schema_version` says; bumping the constant only added a second, redundant way to fail the same record. The bump is reverted and `rejected_count` is now an optional field — the field-set check accepts the full set or the full set minus that key. A v2 record therefore validates exactly as before this PR, and `_valid_owned_process_tree` returns it without inventing the missing key, so the value reads as `None` downstream. Because `None == 0` is `False`, the eligibility bar already treats a missing count as **unknown rather than zero**, with no added branch — the property falls out of the type.

The regression test is deliberately bound to the fleet-level failure rather than the validator in isolation: a healthy, live wrapper is run against a v2-shaped prior through the real wrapped-liveness path, confirming it does not land in `PROCESS_TREE_INVALID`.

**Findings 1, 3 and 5 — closed by one mechanical invariant, not a fourth patched site.** The walk now checks `discovered == admitted + rejected` once after it completes. `discovery_roots` covers the wrapper, declared launcher, discovered brain, and **every prior entry regardless of whether `prior_authority` was trusted** — deliberately, because finding 5's shape is precisely the untrusted case. Each root expands through the current snapshot's raw parent-child edges, and anything reached that is not in `owned` is marked rejected. So a path that drops a candidate can no longer produce a tree with `rejected_count == 0`.

One load-bearing exclusion: a root already proven to be a recycled parent is not expanded, since its current children belong to the replacement process rather than to us. That was not visible in the diff — two pre-existing tests (`..._recycled_launcher_ignores_foreign_children`, `..._recycled_virtual_parent_ignores_replacement_child`) failed on the first pass and are what surfaced it.

Individually: `_snap_index` now excludes a duplicated PID entirely rather than letting the last row win, so lookups correctly report "not found" and the existing `missing_child_row` rejection handles it. The admit/reject disagreement in finding 3 is resolved toward exclusion — a Windows row without exact `start_filetime` is now marked rejected at admission, so the walk and `_valid_owned_process_tree` agree by construction and no tree carrying such an entry can reach `complete`/`absent`. Finding 5's gap was reproduced against `89858e0` first: it reports `status: complete` while a live grandchild is entirely unaccounted — not admitted, not omitted, not rejected. (The test constructs that prior state rather than reaching it through a real multi-poll sequence, and says so.)

**Finding 4** — the validator now rejects `complete`/`absent` with a present, nonzero `rejected_count`, with a control confirming zero still validates.

**Finding 6** — the remedy message gains a clause, emitted only when `rejected_count > 0`, naming rejected candidates as a second class the reset command's identity list (built from `entries`) cannot cover — matching the correction already applied for omitted identities rather than repeating that mistake in the new field.

`test_supervisor.py`: 527 passed, 2 skipped, 1 xfailed, both interpreters. Every new test direction-controlled against `89858e0` individually. Ruff clean. Stop-Tree untouched.

@codex review

zoolok17:
**Disposition for the finding at `ac48478`, addressed in `ca0f5a0`.**

The mechanism was confirmed before the fix: round 5's prior-entry check did `_prior_row = idx.get(pid); if not isinstance(_prior_row, dict): continue`. When `_snap_index` excludes a PID because it appeared in duplicate rows this poll, `idx.get()` returns `None` and that check silently skipped it — treating **"present but ambiguous"** exactly like **"genuinely exited."** Only the second is safe to ignore.

Fixed by splitting `_snap_index` into a thin wrapper over a new `_snap_index_and_excluded`, which also returns the excluded set. The wrapper keeps the original signature and behaviour for its other eight call sites. `_owned_process_tree` now distinguishes the two cases for every prior entry: `idx.get(pid) is None` **and** the PID is in the excluded set → reject; `None` and not excluded → it simply exited, unchanged from before.

The bound matters and is deliberate: this check only ever inspects a PID already recorded as a **prior entry**, so an unrelated agent's unrelated duplicated PID never enters it. That is what makes this closable without reintroducing the failure mode discussed in earlier rounds, where a host-wide malformed-row rule would poison every agent's eligibility for as long as the row kept appearing.

**A test-construction correction worth recording.** The first version of the new test reused the round-4 duplicate-PID fixture shape, with the intermediate's `parent_pid` pointing at the launcher. It passed on the fix — and also passed on `ac48478` **before** the fix, because round 4's discovery closure already reaches a duplicated PID through the permissive edge map whenever it still has a valid edge from a root. That was caught by printing the actual `reason_code` and `rejected_count` rather than trusting the assertion, and the test was rebuilt with an unrelated `parent_pid` so neither `_snap_index` nor the permissive edge map can place it — leaving `prior_entries` seeding it as a root as the only path that reaches it, which is exactly what this fix adds.

That is the second false-confirming green caught this way on this PR. A test that passes for the wrong reason is worse than no test, and essentially nothing surfaces it except reading the real values.

`test_supervisor.py`: 533 passed, 2 skipped, 1 xfailed, both interpreters. New test direction-controlled against `ac48478` with the corrected fixture. Ruff clean. Stop-Tree untouched.

Findings by round on this PR: 4 → 6 → 3 → 2 → 1 → 0 pending this review.

Follow-up carved out rather than folded in: #109, plumbing the launch barrier's observed survivors through as counted rejections instead of an unknown. Deliberately diagnostics rather than safety — the "missing means unknown, never zero" fix already closes the vulnerability.

@codex review

zoolok17:
**Disposition for the finding at `ca0f5a0`, addressed in `3d8c2d6` — and the class is now closed structurally.**

The named regression is fixed: `_ephemeral_process_alive` returns `None` (unknown) rather than `False` for an excluded launcher PID, and `_plan_ephemeral_active` already treats only `alive is False` as the terminal-failure signal, so no downstream change was needed. The test asserts the **outcome** an operator would see — `action != ACTION_FAILED` and `held_terminal` unset — not the internal flag, and direction-controls against `ca0f5a0` with the exact `terminal_state: failed` it previously produced.

**The survey of all seven other consumers produced three different outcomes, and two of them were corrections to the author's own first attempt.**

Two sites (`_unverified_owned_process_tree`'s promotion check, and `evaluate_launch_barrier`) were initially "fixed" — and both fixes passed *identically* on the pre-fix commit. Tracing why: both gate the excluded-set-consuming code behind `snapshot_error is None`, and `_snap_index`'s exclusion criterion (a valid PID on 2+ rows) is a strict subset of what `_snapshot_pid_integrity_error` already detects over the same snapshot. So at those points the excluded set is provably empty and the fixes were unreachable code. Both were **reverted** and replaced with a comment explaining why the sites are safe as written.

One site (`_attribution`'s launcher confirmation) is genuinely indifferent: it only suppresses kill-targeting for an ambiguous identity, which is already this file's stated default.

Three sites (`_unverified_owned_process_may_exist`, the wrapper-liveness check, and legacy brain liveness) were fixed on principle, but **no discriminating end-to-end test could be constructed** — in each case a pre-existing safety net produces the same safe outcome first. Rather than ship tests that pass for the wrong reason, those tests were removed and honest comments left at all three sites recording exactly that: fixed because the rule is correct and cheap, not proven to currently matter, and not proven dead the way the two reverted sites were.

**The ergonomic closure landed.** With all eight sites migrated to `_snap_index_and_excluded` — using the excluded set or explicitly discarding it with a stated reason — the thin `_snap_index` wrapper had zero callers and was deleted. A future caller can no longer obtain an index without also being handed the set whose omission caused this bug twice in two rounds.

`test_supervisor.py`: 534 passed, 2 skipped, 1 xfailed, both interpreters. Ruff clean. Stop-Tree untouched.

Findings by round: 4 → 6 → 3 → 2 → 1 → 1 → 0 pending this review.

@codex review

zoolok17:
**Disposition for the two findings at `3d8c2d6` (one defect, both ends), addressed in `d8c5000`.**

The round-7 survey classification of `_attribution` as "genuinely indifferent" is **retracted**, and the PR description's round-7 summary bullet was corrected to point at this section rather than leaving a withdrawn claim standing.

Mechanism confirmed as reported: the kill-targeting loop and the provenance loop both do `idx.get(entry.get("pid"))` on `valid_priors` entries and silently drop the entry on a miss — so *genuinely absent* and *merely excluded* read identically, exactly as everywhere else this round, but applied to a **prior-tracked descendant** rather than the launcher the survey was looking at.

**Kill-targeting**: an excluded entry now yields a synthetic row built from the prior record's own identity (`pid` plus `start`), which is the only identity this attribution model ever carried for such entries — they have no `start_filetime` at all, confirmed against `_new_provenance_entry`'s schema. It is recorded through the same `_record_target` path as every other target, so `Stop-Tree`'s own start-time guard remains what decides whether it is killed.

**Provenance**: an excluded entry is retained unchanged in `managed_by_pid` rather than omitted, so the next poll's `valid_priors` still carries it once the ambiguity clears.

**A test-discrimination wrinkle worth recording.** The kill-targeting half was verified on a manual-restart poll and direction-controls against `3d8c2d6` (`kill_target_pids == set()` before the fix). That same poll **cannot** verify the provenance half, because a manual restart legitimately starts `managed_pids` fresh on relaunch regardless of this change — so neither outcome there discriminates anything. A second poll was added (fresh heartbeat, no restart marker, same duplicated-PID snapshot) and the provenance half was verified **in isolation against the pre-fix commit first** — empty before, containing the PID after — before the combined test's result was trusted.

**Stated residual, flagged rather than built:** the BFS descendant-walk immediately below the two fixed loops has the identical `idx.get(parent_pid) is None` pattern, gating whether a tracked seed's children are walked. Closing it is not a copy of the same one-line fix — `_strict_child_edge` needs a real parent row for start-time ordering, and an excluded seed has no fresh row to offer, so it would require a variant of that edge check tolerant of a parent it cannot fully verify. That is a separate and larger change; it will be filed as its own issue rather than folded in.

The methodology correction was also applied backwards over the earlier survey: the two reverted sites were re-traced under "what do downstream consumers do with this output" rather than re-asserted — one is a pass-through of an already-correct prior tree, the other is consumed once within the same poll and never persisted.

`test_supervisor.py`: 535 passed, 2 skipped, 1 xfailed, both interpreters. Ruff clean. Stop-Tree untouched.

@codex review

zoolok17:
**Disposition for the three findings at `d8c5000`, addressed in `070d8dd`.**

**1. The double-count — fixed, and the stale rationale retired with it.** An unproven virtual-parent descendant was marked rejected once in the main walk and again by the discovery-closure loop, since it is never admitted into `owned`. `mark_rejected` now takes an optional PID and dedups through a `rejected_pids` set; all 19 call sites with an identifiable single candidate pass it (the closure loop's own PID was previously discarded via `for _ in ...` and is now used), while sites with no single candidate correctly pass none.

The important part is what followed. This defect existed because an earlier round justified double-counting as "harmless — `rejected_count` is a gate, not a tally", which was true when written and stopped being true when a later round added an operator-facing message that reports the number. Every remaining comment invoking that same justification was re-read: **two more sites were relying on it**, both were fixed, and the stale rationale language was deleted rather than left sitting next to code that no longer needs it. Dedup now handles it structurally instead of by argument.

Test asserts the operator-visible string — "excluded 1 candidate(s)" — with the counter checked only for context. Direction-controlled: `rejected_count == 2` on `d8c5000`.

**2. The refusing remedy — closed as a class, not as an instance.** This was the third message on this PR naming a command the CLI would reject. The configured-agent reset path reads the current wrapper runtime record fresh and refuses unless it is `STATUS_VALID`, independent of the tree's entries or rejection count. The remedy now gates on `liveness["runtime_status"] == STATUS_VALID` — **the same signal `_wrapped_liveness` already computed this poll, over the identical record the command re-reads at reset time**. When it is not valid the command name may still appear for context, but the *recommendation to run it* does not; the message points at the reason code instead, matching the wording pattern already approved for entryless placeholders.

Stated honestly rather than papered over: this is a plan-time read, and the runtime record could change between the poll and the operator acting. That gap cannot be closed from where the message is produced, and it is the same staleness every HOLD message already carries — not something introduced here.

**3. The malformed-PID variant — half was already safe, half is a documented residual.** Both sub-cases were built and executed rather than reasoned about.

For a row that was *never* a prior entry, with a live child below it: `rejected_count == 0` and both identities absent from entries — the reported defect, confirmed.

For a *prior-tracked* intermediate with an identical malformed-PID row: both the intermediate and its live child end up correctly in `entries`. The trusted-rehydration path never reads this poll's row for a needed ancestor — it reconstructs identity from the prior record's own `start`/`start_filetime` and re-verifies the live child against that same chain, bypassing the malformed row entirely. Pre-existing machinery, not added here.

So the bound holds exactly where predicted: a row whose *intended* identity matches a prior entry is already safe, because the rehydration path never needs the corrupted field. For a row with no prior entry to anchor it, identity is precisely what is missing — every other fix in this family works by looking a row up *by* its intact PID, and no snapshot-only signal substitutes when the PID itself is corrupt. Accepted as a **documented residual**, recorded in `_snap_all_edges`'s own comment, with two tests: one locking in the safe case as a genuine regression guard, one demonstrating the residual by execution. Both behave identically on `d8c5000` — characterization, not fix, and labelled as such.

`test_supervisor.py`: 539 passed, 2 skipped, 1 xfailed, both interpreters. Ruff clean. Stop-Tree untouched.

@codex review

zoolok17:
**Disposition for the two findings at `070d8dd`, addressed in `c05a2b1`.**

**1. The deferred BFS gap — built, after resolving the design question rather than deferring it again.** `_children_map(idx)` is keyed by each row's own valid `parent_pid`, so it can list an excluded seed's children without the seed's row being present — but the BFS loop's `parent = idx.get(parent_pid); if not isinstance(parent, dict): continue` short-circuits before `kids` is ever consulted, the identical excluded-PID miss fixed one loop earlier in round 8.

The open question was whether a synthetic parent row is sufficient for `_strict_child_edge`. It reads exactly two fields off `parent` — `parent["pid"]` and `parent["start_time"]` — and round 8's synthetic row already carries both; `Stop-Tree`'s own start-time guard already trusts that same row to order this PID's kill. So sufficiency came back **yes**, and no "unaccounted" branch was needed at this site. The fallback is scoped strictly to the excluded-seed case: every other PID reaching the queue got there after `idx.get(child_pid)` was already confirmed a dict, so a seed with neither a real nor a synthetic row is unreachable by construction — the `continue` remains as a documented backstop, not a live branch.

**2. The fourth refusing remedy — closed with one shared function, not a second approximation.** `process_tree_ownership_reset_evidence` separately requires `entry["launcher_pid"] == wrapper_pid` plus a start-token match, and `_wrapped_liveness` runs the *identical* comparison at its `wrapper_state_mismatch` branch without ever surfacing the result. A record can be `STATUS_VALID` — it parses — while still failing this, which is exactly the case the previous predicate could not see.

Factored into `_wrapper_identity_matches_state(record, st)` with **three** call sites pointed at it: `_wrapped_liveness`'s mismatch branch (previously the loosest of the three — no type or positivity guards, now matching the others), the command's own admission check, and the remedy predicate. Not a re-derivation that agrees today; the same function, fed the record `_wrapped_liveness` already normalised this poll.

Also split the `has_entries` wording, because one predicate failing for two different reasons was about to emit a literal contradiction — "runtime record is not valid (valid)" — in the mismatch case. It now names which condition the command will actually refuse on.

**A test-integrity catch that matters beyond this round.** The first run of the new BFS test appeared to pass on *both* the fixed and the stashed commit. Rather than concluding the fix was inert, `sup.__file__` was checked — and pytest was importing the **main checkout's installed `supervisor.py`** via the editable install, not this worktree's. `git stash` in a worktree has no effect on that. Everything was re-run with `PYTHONPATH` pointed at the worktree's `src`, and the failure/pass split is real.

That is worth stating plainly: any direction-control claim made without that environment pinned proves nothing about the worktree's code. It was caught only because "passes either way" matched the false-confirming-green pattern closely enough to prompt a check rather than a conclusion.

`test_supervisor.py`: 541 passed, 2 skipped, 1 xfailed, both interpreters, both runs with `PYTHONPATH` pinned. Ruff clean. Stop-Tree untouched.

Findings by round: 4 → 6 → 3 → 2 → 1 → 1 → 0 → 2 → 3 → 2 → **0 pending this review**.

@codex review

zoolok17:
**Disposition for the two findings at `c05a2b1`, addressed in `c74f776` — and the structural question is closed with a decision.**

**1. The stale zero — fixed and swept.** `_current_proof_failed`'s only path into a re-derivation attempt runs through the held fallthrough: a `copy.deepcopy(prior)` relabelled to `invalid`, with no walk performed. It was carrying `prior`'s `rejected_count` forward as if it described *this* conversion, which never asked the question. The field is now popped — the Python-side analogue of round 4's PowerShell precedent, same rule: **a producer that does not walk must not claim to have walked.**

Swept the rest of the same-shaped sites rather than stopping at the reported one. Three passthroughs return `prior` unchanged and relabel nothing, so they claim nothing new. `_invalid_owned_process_tree_record` hardcodes `rejected_count: 0` but pairs it with `entries: []`, and the gate requires `bool(entries)` — verified by reading the gate rather than pattern-matching the shape. The sibling *absent*-promotion branch was left unchanged after tracing every reader: eligibility never consults the count once status is `absent`, no remedy path reads it there, and the validator accepts `0` or missing. Proven inert, not assumed safe by analogy.

**2. Excluded child of a trusted root — fixed, deliberately NOT the way rounds 8 and 10 were.** This is round 10's shape inverted: trusted parent, excluded *child*, undiscoverable because `_children_map` is built from `idx.items()` and an excluded child contributes no edge.

The synthetic-row treatment was **not** extended here, and the reasoning is the substance: rounds 8 and 10 built a synthetic row from an *independent, already-trusted source* — the prior record, earned on an earlier clean poll. A first-time-seen child has no such record; the only information about it is the same ambiguous duplicate rows that caused its exclusion. Synthesising from those would mean trusting an identity from exactly the evidence `_snap_index_and_excluded` exists to distrust — structurally the documented `_snap_all_edges` residual, reached by row duplication instead of a malformed field.

So it is neither targeted nor persisted, but no longer silently lost: the BFS now walks the permissive edge map per visited parent and bumps a deduped diagnostic counter for any excluded child found there. (`_bump` is a no-op against an unregistered key, which produced silently zero counts on the first run until the counter was registered.)

**The structural question — closed, with a decision to stop.**

Counted before deciding: **42 accessor touch points** on `idx`-shaped data across the file, against **8 sites** that have ever needed the excluded-versus-absent distinction to reach a correct decision; the remainder are indifferent by design or already paired with an explicit check.

A discriminated-result API would require migrating all 42 to stay coherent, including ~34 already-correct ones — and Python offers no enforced exhaustiveness, so it would make the excluded case *visible* everywhere without picking the *right* handling anywhere. And the right handling has genuinely differed by site: a trusted synthetic row where an independent anchor exists (rounds 8, 10), a documented residual where none exists and the PID is unreadable (round 9), a diagnostic-only bump where none exists but the PID is readable (round 11). Three distinct, each-correct resolutions turning on domain facts no type can see.

**A structural fix would surface the fork; it cannot resolve it.** The per-site convention this file already uses — proving indifference or necessity at each site rather than asserting it — is doing the work a discriminated type would claim credit for, at a fraction of the churn. Accepted as **enumerated-and-fixed with a bounded, judged residual**, and the offered partial rename declined on the same reasoning: it touches correct sites for no behavioural gain.

`test_supervisor.py`: 543 passed, 2 skipped, 1 xfailed, both interpreters, `PYTHONPATH` pinned throughout. Both findings direction-controlled against `c05a2b1`. Ruff clean. Stop-Tree untouched.

@codex review

zoolok17:
**Disposition for the three findings at `c74f776`, two fixed in `32668ee` and one deliberately deferred into a structural change.**

**1. Unaccounted excluded child now HOLDs instead of incrementing a counter.** The previous round added a diagnostic that nothing gated on, so relaunch still proceeded with an unaccounted live child — a check that runs and is never read is not a check. The HOLD now sits at the same precedence as the wrapped path's own `tree_status` check, *before* the manual-restart marker is read, so it blocks that path unconditionally rather than only the automatic one. Returns `WARN_ONLY`/`UNACCOUNTED_LIVE_DESCENDANT` with empty `kill_targets` — nothing is killed at all while the ambiguity stands, on the same reasoning as a partial wrapped tree.

**2. The recycled-seed traversal — reverted rather than patched.** The mechanism is worth stating precisely: the synthetic row is safe as a *kill target*, because `Stop-Tree` re-verifies the live process's actual start time before acting and refuses a recycled PID. It is **not** safe as a `_strict_child_edge` **parent**, because nothing re-verifies the time-ordering claim before it feeds an independent kill decision about a *different* PID — and a recycled PID's stale prior start time is almost always earlier than any of the replacement's children, so the ordering check meant to reject unrelated descendants would pass every one of them.

No recycling detector was added, because snapshot-only evidence cannot supply one — the same "no signal to disambiguate" shape already documented elsewhere. Instead round 10's BFS reuse is reverted outright: an excluded parent is no longer descended through via a synthetic identity at all, and its children now take the same "unaccounted" path built for an excluded *child*. That converges three rounds of overlapping rules into one — and is only safe to converge because the HOLD above landed first; without it, "unaccounted" would be a quieter version of the same hazard.

Both existing BFS tests were updated in place to assert the corrected outcome and direction-controlled; the round-10 test fails on the old code with `action == 'relaunch'`, confirming the hazard was live rather than theoretical.

**3. The narrowed-rebuild case — deferred into the inversion, not patched as a fifth instance.** `prior_record_invalid` sets `invalid_reason` upfront and discards `prior_by_pid`, so the walk's discovery roots never include the historical tree. If the narrowed set admits cleanly, `rejected_count` stays 0 and completeness reads true — **honestly computed for the graph this walk saw, and wrong for the graph that matters, which the walk never had a chance to check.**

**The structural change, sized and scheduled for the next round.**

Every finding from rounds 9 through 12 is one sentence: a record was produced without a complete walk, and `rejected_count == 0` came along for the ride — inherited, defaulted, hardcoded, or untouched by a producer that only relabelled a status. Fixing each producer chases *absence of counter-evidence*, which cannot converge, because zero is an integer's default and every new path gets it free.

The inversion: a `walk_complete` boolean, optional in the schema (missing reads as unknown, already ineligible — the same migration-safe shape `rejected_count` uses), set in exactly one place at the end of the walk as `bool(entries) and omitted_count == 0 and rejected_count == 0 and not prior_record_invalid`.

That is the **identical conjunction** completeness already computes today. The change is not *what* is checked but *where*: computed once at the source while the inputs are freshly known, instead of re-derived by every later reader from fields that can be inherited or defaulted without ever having been jointly checked. Counts remain accurate for the operator-facing message but stop being independently load-bearing for the safety decision.

On whether a legitimate path must promote to absent without a walk this poll: yes — a wrapper confirmed absent from a current snapshot, checked against a prior record. That is not a carve-out. `walk_complete` is a persistent fact about the *stored* record, earned on whichever earlier poll actually walked, exactly as completeness is read from the prior today rather than recomputed. The inversion changes representation, not timing. `prior_record_invalid` and "no walk ran" are excluded by the formula's construction rather than as special cases.

Four touch points, one cross-language (the PowerShell post-kill barrier gains a sibling property removal). Deliberately **not** built this round: two behaviour-changing fixes already landed here and deserve their own scrutiny before a third, larger one stacks on top.

`test_supervisor.py`: full suite, both interpreters, `PYTHONPATH` pinned. Stop-Tree untouched.

@codex review

zoolok17:
**Disposition for the TTL finding at `32668ee`, addressed in `fadea84`.**

Round 8 kept an excluded prior entry in `managed_by_pid` unchanged as ambiguous HOLD evidence — but "unchanged" meant its `last_fresh_attribution_epoch` never moved either, so `_prior_valid`'s TTL check compared a frozen value against `now_epoch` every poll and eventually dropped it after an hour of continuous, never-resolving ambiguity. For a seed whose `parent_pid` is unrelated to the launcher, that entry is the only root discovering its opaque child, so it took the `UNACCOUNTED_LIVE_DESCENDANT` HOLD down with it.

Fixed by refreshing the epoch in the excluded branch — treating "still ambiguous" as a fresh observation rather than an aging one.

**A wrong turn, reported rather than smoothed over.** The first attempt refreshed the epoch in *both* branches, on the reasoning that neither is re-derived through `_new_provenance_entry` since `prior_only` is set unconditionally for both. That reasoning was correct as far as it went, and it broke a test that deliberately pins a valid match's epoch *not* moving. The missed distinction: a validly-matching identity has other paths to earn a fresh stamp — `own_wrapper`, `own_wait`, `live_chain_descendant`, `legacy_rederived`, all of which need a real index row, which it has. An excluded identity has none of those by construction while the ambiguity persists. The full suite surfaced it before the reasoning did; both sides are now documented at the call site so the next round need not re-derive them.

**Survey — only this one, with two useful negatives.** `_PROVENANCE_TTL_SECONDS` is the sole safety-gating TTL in the file. The launch-coordination barrier's own `last_observed_epoch` *already* refreshes on every poll the same blocking condition is observed — the correct pattern is live elsewhere in this file, so this fix matches an established shape rather than inventing one. And the heartbeat/progress-age staleness checks are genuinely decaying evidence — nobody has heard from the process — correctly TTL'd as-is, not another instance.

**Coupling with the planned inversion — none, traced by dispatch rather than by shape.** `_liveness` routes each agent to exactly one of `_wrapped_liveness` or `_attribution` on the `wrapped` flag, never both. `_prior_valid` and the TTL live only inside `_attribution`'s `managed_pids` handling; `walk_complete` would live only inside the owned-process-tree handling. Disjoint structures, disjoint paths, no agent runs both.

Test: three chained `plan_actions` calls 2000s apart with the ambiguity held constant, so the cumulative 4000s gap exceeds the one-hour TTL while no individual gap does. Asserts the operator-visible outcome on the third poll — still `WARN_ONLY`/`UNACCOUNTED_LIVE_DESCENDANT` with empty `kill_targets`. Direction-controlled against `32668ee`, and the pinned valid-match test was separately re-run against that same pre-fix commit to confirm the revert did not paper over a second bug.

`test_supervisor.py`: 544 passed, 2 skipped, 1 xfailed, both interpreters, `PYTHONPATH` pinned. Ruff clean. Stop-Tree untouched.

@codex review

PR #110, opened 2026-08-02
walk_complete: positive-evidence eligibility for owned-process-tree re-derivation

zoolok17:
## Summary

Follow-up to #108 (task #150). Rounds 9-13 of that PR each found a different producer that could mark an `owned_process_tree` record eligible for absence-promotion without a real walk ever running: an inherited `rejected_count`, a hardcoded placeholder default, a status relabel, a narrowed rebuild over a discarded prior. Every fix closed one instance by making that specific producer stop claiming zero - chasing the ABSENCE of recorded rejections, which is the default state of an integer and therefore free to any new or narrowed code path that never ran a walk. The lead's own framing: the count never converged because the instrument was wrong, not the diligence.

This inverts the instrument. `walk_complete` is positive evidence a complete walk actually ran, set in exactly one place - the end of `_owned_process_tree`'s own walk - as the identical conjunction `_unverified_owned_process_tree` used to re-derive as `entries_are_complete` (`bool(entries) and omitted_count == 0 and rejected_count == 0`), plus `not prior_record_invalid` (closing the narrowed-rebuild gap that conjunction alone could never close, since `prior_record_invalid` is local to the call that produces the record). Computed once, at the source, while every input is freshly known - never re-derived by a later reader from fields that can be inherited, defaulted, or hardcoded without ever having been jointly checked.

`rejected_count`/`omitted_count` remain in the schema as diagnostics and the operator-facing remedy-message count, where accuracy still matters, but they stop being independently load-bearing for the safety decision.

## Touch points (traced by consumer, not just producer)

- `_owned_process_tree`: sets `walk_complete` once, at the walk's own end.
- `_unverified_owned_process_tree`: eligibility reads `walk_complete` instead of re-deriving `entries_are_complete`; the "held" (unwalked-relabel) conversion strips `walk_complete` the same way it already strips `rejected_count`.
- `_valid_owned_process_tree` (schema/validator): `walk_complete` is the second optional field, independently present/absent from `rejected_count` (a record may predate one, the other, both, or neither); a complete/absent record explicitly claiming `walk_complete=False` is flagged as internally contradictory, mirroring the existing nonzero-`rejected_count` check.
- The PowerShell post-kill/ephemeral-teardown barrier: gains a sibling `PSObject.Properties.Remove('walk_complete')` at both mutation sites, next to the existing `rejected_count` removal.

Surveyed and confirmed unaffected: `_plan_one`'s remedy message (rejected_count stays diagnostics-only, its own branch gating never depended on it) and `evaluate_launch_barrier` (only ever reads the final status classification, never the counts underneath it).

## Test plan
- [x] Updated two existing tests whose fixtures manually deleted only `rejected_count` from a real walked tree to simulate a pre-migration record - the tree now also carries `walk_complete: True` (honestly earned), so deleting only one field no longer represents "before this feature existed"; one test now demonstrates the inversion's own defining property instead (rejected_count alone is no longer load-bearing for eligibility).
- [x] Added a direct regression test for the narrowed-rebuild hazard (a malformed prior with a live wrapper rebuilds a narrower graph that read complete under the old scheme), direction-controlled against c92b019: fails on the pre-inversion code (missing `walk_complete` field), passes with the fix; independently confirmed by execution that the pre-inversion code's later-poll check actually promotes to `absent` (the real hazard), not just that the field is missing.
- [x] Full `tests/test_supervisor.py` suite green on the default interpreter and `py -3.10`: 545 passed, 2 skipped, 1 xfailed (543 baseline + 2 existing-test bodies updated, no net new failures). `PYTHONPATH` pinned from the first test run of the session, per the standing lesson from the prior PR.
- [x] `ruff check` clean on both touched files.
- [ ] dev-gate/build/wheel/security legs intentionally not run locally (targeted tests only, per standing directive) - left to CI.

Did not touch Stop-Tree.

### DIFF
--- a/src/agenttalk/supervisor.py
+++ b/src/agenttalk/supervisor.py
@@ -2675,6 +2675,25 @@ def _valid_owned_process_tree(
         or value.get("truncated") is not (omitted_count > 0)
     ):
         return None
+    # Task #150 follow-up connector finding: the checks above only typed
+    # walk_complete and rejected the False-with-complete/absent
+    # contradiction - they never constrained True against the conjunction
+    # that is supposed to have earned it. A persisted record claiming
+    # walk_complete True while omitted_count/rejected_count are nonzero, or
+    # rejected_count is unknown, or entries is empty, is exactly as
+    # internally contradictory as the existing nonzero-rejected_count-on-
+    # complete/absent check above (same shape, same round-3 precedent) -
+    # and reading it as authoritative here would make the inversion WORSE
+    # than the re-derivation it replaced: the old entries_are_complete
+    # re-derivation would have caught this record; a bare type-check on a
+    # single boolean would not. This does not require the flag to be
+    # PRESENT - only present-and-True must match its own preconditions;
+    # absent or False stays eligible-by-nothing (unknown), never rejected
+    # for a conjunction the record never claimed to have earned.
+    if walk_complete is True and not (
+        bool(entries) and omitted_count == 0 and rejected_count == 0
+    ):
+        return None
     reason_code = value.get("reason_code")
     if value.get("status") == "complete":
         if reason_code is not None or omitted_count:
--- a/src/agenttalk/supervisor.py
+++ b/src/agenttalk/supervisor.py
@@ -197,7 +197,28 @@ _SAFE_WRAPPER_GENERATION_RE = re.compile(
 # record never earned. schema_version is deliberately NOT bumped: the
 # field's own optionality is the migration, so there is nothing for a
 # version bump to gate that isn't already handled without one.
-_OWNED_PROCESS_TREE_OPTIONAL_FIELDS = frozenset({"rejected_count"})
+#
+# The walk_complete inversion (task #150 follow-up): rejected_count == 0
+# (with omitted_count == 0 and at least one entry) was the eligibility bar
+# for re-deriving absence from an invalid prior - but that bar is chasing
+# the ABSENCE of recorded rejections, which is the default state of an
+# integer and therefore free to every new or narrowed code path that never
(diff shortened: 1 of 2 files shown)


zoolok17:
**Disposition for the P1 at `7378063`, fixed in `56d5191`.**

The finding was correct and the mechanism it named was live: the validator checked only that `walk_complete` was a boolean, so a corrupted or partially-updated record could carry `walk_complete: true` alongside `omitted_count > 0`, `rejected_count > 0`, or empty `entries`. `_unverified_owned_process_tree` reads the boolean alone as eligibility, so such a record could be promoted to `absent` once its recorded subset disappeared — authorizing teardown or relaunch while omitted or rejected owned processes may still be live.

`56d5191` makes the validator enforce the same conjunction the production site computes: `walk_complete is True` now requires nonempty `entries`, `omitted_count == 0`, and `rejected_count == 0`. A record that claims completeness it cannot substantiate is rejected at read time rather than trusted.

That is the whole point of the inversion this PR exists for, applied to its own read path. The original defect class was that `rejected_count == 0` arrived by inheritance or default and was believed; a `walk_complete: true` that arrives the same way and is believed would be the identical bug wearing the new field's name. Constraining the boolean against its own inputs is what stops the new representation from reintroducing what it replaced.

Field note, not a change request: this PR does not address the case measured on our own fleet today, where an `invalid` tree with a *healthy* wrapper is never re-walked at all — `_owned_process_tree` returns the stored record before the current snapshot is parsed, so no eligibility field of any kind is consulted. That is tracked separately and is a design call for its own round. Flagging it only so a reviewer does not assume this PR closes it.

Full `test_supervisor.py`, both interpreters, `PYTHONPATH` pinned.

@codex review

PR #111, opened 2026-08-04
BLOCKED (redesign): fix(#163 launch half): additive absence proof for evaluate_launch_barrier

zoolok17:
## Summary
- Second half of the #163 launch-half fix (planner side already merged directly to master in 7de07e4, per an instruction gap now corrected - this half goes through a PR).
- `evaluate_launch_barrier`'s primary generation-match stays exactly as strict as it is today; it gains an additive, independent second sufficient condition (the ALTERNATIVE PATH) that proves absence directly from the current snapshot via the identical shared predicate the planner side already uses, rather than depending on the generation-match proxy in the not-yet-adopted state where that proxy is wrong.
- `_no_live_identity_for_launch` / `_named_identities_confirmed_absent` / `_claimed_self_identities` are now shared across all three call sites (two in `_wrapped_liveness`, one new in `evaluate_launch_barrier`) - one implementation, not three.
- `cli.py`'s `--launch-barrier` handler now reads a fresh wrapper runtime record and threads it through as the barrier's `runtime_view` parameter.

## Test plan
- [x] `tests/test_supervisor.py` + `tests/test_coordination_stall.py`: 582 passed, 2 skipped, 1 xfailed (this interpreter, `PYTHONPATH` pinned to this worktree's `src`, run from the worktree cwd).
- [x] Four new direct tests on `evaluate_launch_barrier`: flagship generation-mismatch + empty snapshot now allows the spawn; the same shape with the claimed identity alive in the snapshot still blocks (direction control); a schema-drifted/corrupted prior is excluded even with an empty snapshot; legacy evidence still blocks on an unrelated live row elsewhere (mirrors the planner-side deep-orphan test).
- [x] Confirmed red-first by direct probe before this change: same tree/call, stale generation -> validator rejects -> barrier blocks; generation dropped -> validates.
- [ ] NOT YET DONE: the red-first fleet-level test with the external effect oracle (exactly one replacement process, reaches readiness, exact identity, no duplicate/no kill, exact cleanup). The supervisor's own spawn step has no pure-Python path - it's `Start-Process`/`CreateProcessW` inside the generated PowerShell executor, a different and heavier shape than the wrapper-level stub-agent canary infra. Flagging this gap rather than building new PowerShell-driving test infra under time pressure in the same PR.

## Reviewer note
Two test assertions in the prior commit (7de07e4, already on master) were updated to reflect intentionally-changed behavior rather than reverted - `test_owned_process_tree_generation_adoption_reaches_recovery_without_visible_roots` and the final scenario of `test_attended_process_tree_reset_retires_stale_runtime_before_relaunch_plan`. Flagging both by name per the operator's request, since an edited expectation is not independent evidence on its own.

### DIFF
--- a/src/agenttalk/cli.py
+++ b/src/agenttalk/cli.py
@@ -12509,6 +12509,13 @@ def cmd_supervise(args: argparse.Namespace) -> int:
         state = _read_state()
         config = _load_supervisor_config(store)
         now = args.now if args.now is not None else time.time()
+        # #163 launch half: a fresh read, mirroring how build_report seeds
+        # the planner's own wrapper_runtime - the barrier's alternative
+        # absence path needs the CURRENT runtime record's own wrapper/CLI
+        # launcher identity, not a proxy for it.
+        barrier_runtime_view = runtime_obs.read_runtime(
+            store.state_dir, args.agent, now_epoch=now,
+        )
         result = sup.evaluate_launch_barrier(
             _read_snapshot_file(args.snapshot_file),
             state,
@@ -12516,6 +12523,7 @@ def cmd_supervise(args: argparse.Namespace) -> int:
             args.agent,
             root_key=sup._root_key(str(store.root.resolve())),
             request_id=args.request_id,
+            runtime_view=barrier_runtime_view,
         )
         if result.get("blocked") and getattr(args, "record_events", False):
             with contextlib.suppress(Exception):
--- a/src/agenttalk/supervisor.py
+++ b/src/agenttalk/supervisor.py
@@ -2382,33 +2382,98 @@ def _named_identities_confirmed_absent(
     return True
 
 
+def _claimed_self_identities(
+    legacy_process_evidence: object,
+    runtime_record: object,
+) -> tuple[list[tuple[object, object, object]], list[tuple[object, object, object]]]:
+    """Every identity this agent's own evidence currently claims as its
+    own, split into ``(legacy_identities, other_identities)``. Legacy
+    pre-v2 candidates (wrapper/brain/managed rows, already deduplicated by
+    ``_legacy_process_migration_evidence``) come first, since they carry no
+    bounded parent graph and callers must apply a stricter bar to them; the
+    runtime record's wrapper pid+start and, when a CLI launcher is expected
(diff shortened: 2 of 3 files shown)


zoolok17:
BLOCKED on a #163 redesign - do not merge.

7de07e4 (the planner-side half this PR builds on) has been reverted from master (627d5ee). reviewer-1's cold review found a reproduced, structural blocker in the shared predicate this PR's alternative path also depends on (`_no_live_identity_for_launch` / `_named_identities_confirmed_absent` / `_claimed_self_identities`):

- `_claimed_self_identities` only includes the runtime record's CLI-launcher identity while `phase == ACTIVE`. A validated TERMINAL record still carries `cli_launcher_pid`/`cli_launcher_start`, but those get dropped from the checked set.
- `_named_identities_confirmed_absent` only excludes rows whose CURRENT DIRECT `parent_pid` is a named root - it cannot see a launcher/tool reparented after its recorded parent exited, or a grandchild whose intermediate parent exited.

Reproduced directly: same state, absent wrapper, snapshot containing the runtime record's exact live CLI-launcher PID/start reparented to PID 1 - `phase=ACTIVE` gives `warn_only` (correct), changing only `phase` to `TERMINAL` gives `stuck_recover`/`STUCK_OR_DEAD` (a false absence proof, while that exact process is alive). A second probe reproduced the same false absence for a live orphaned grandchild.

The structural problem: the predicate proves absence by enumerating identities it knows about, and the failure mode is the ones it doesn't. Adding TERMINAL to the phase check does not fix this - the claimed universe stays incomplete for a wrapper visible before its first runtime write, and for any descendant whose current parent edge no longer reaches a named root.

Not a criticism of this PR's own construction - the alternative-path SHAPE here (primary generation-match untouched, re-validate with generation dropped, prove absence, then run the barrier's own survivor scan unchanged) is expected to be the right shape once the proof underneath it is sound. Holding until the redesign lands.

zoolok17:
Closing on operator approval of the 2026-09-17 backlog triage: self-declared BLOCKED (redesign) since 2026-08-04 and untouched since. The planner-side half is on master; reopen the launch-barrier half from a new design if the need returns.

PR #112, opened 2026-08-04
MOOT (superseded by revert): hotfix(#163): fix unbound prior_record_invalid crash

zoolok17:
## Summary
- Master went red immediately after 7de07e4: `tests/test_ephemeral_reviewers.py::test_launched_exit_without_fresh_tree_holds_without_restart_or_archive` fails on every platform.
- Root cause is a `NameError`, not the cross-module semantic conflict it looked like from the test name: `prior_record_invalid` is only assigned inside `if isinstance(root_key, str):` in `_wrapped_liveness`, but `_no_live_identity_here` (called unconditionally from `_current_proof_failed`) reads it regardless. The ephemeral-reviewer path (`_ephemeral_owned_process_view`) calls `_wrapped_liveness` with `root_key=None` for a launched reviewer with no fresh runtime record - a combination the wrapped-agent path never produces, since `plan_actions` always normalizes `root_key` to a string first.
- Fix: initialize `prior_record_invalid = False` alongside the existing `prior_tree = None` / `raw_prior = None` pre-initialization.
- Confirmed the named test passes with its ORIGINAL, unmodified assertions once this one line is fixed - it was never touched.

## Caller audit (requested)
- `_wrapped_liveness` has exactly two callers: `_liveness` (feeds `_plan_one`) and `_ephemeral_owned_process_view` (feeds `_plan_ephemeral_active`).
- `_plan_ephemeral_active` computes `no_live_identity_for_launch` as a side effect of flowing through `_current_proof_failed`, but never reads that field for its own decision - so its behavior is unchanged, not coincidentally passing.
- `evaluate_launch_barrier` has exactly one caller (`cli.py --launch-barrier`).
- No fourth call site exists.

## Test plan
- [x] `tests/test_ephemeral_reviewers.py` + `tests/test_supervisor.py` + `tests/test_coordination_stall.py`: 611 passed, 2 skipped, 1 xfailed.

### DIFF
--- a/src/agenttalk/supervisor.py
+++ b/src/agenttalk/supervisor.py
@@ -5661,6 +5661,15 @@ def _wrapped_liveness(
     diagnostics = _diag()
     prior_tree = None
     raw_prior = None
+    # #163 hotfix: only ever reassigned inside `if isinstance(root_key, str):`
+    # below, same as prior_tree/raw_prior above it - but _no_live_identity_here
+    # (called unconditionally, from _current_proof_failed, regardless of
+    # root_key) reads it too. A non-string root_key (the ephemeral-reviewer
+    # caller passes one) skipped that block entirely and left this name
+    # unbound - a crash, not a scoping decision. False here means exactly
+    # what raw_prior=None already means for that case: no prior record to
+    # call invalid.
+    prior_record_invalid = False
     legacy_evidence = _legacy_process_migration_evidence(st)
     prior_generation = st.get("runtime_wrapper_generation")
     has_revoked_runtime = "revoked_wrapper_runtime" in st


zoolok17:
MOOT - superseded by the revert. 7de07e4 (the code this PR patches) has been reverted from master (627d5ee) after reviewer-1's cold review found a separate, more fundamental soundness blocker in the shared predicate (see PR #111's pinned comment for the details - a false-absence proof for reparented/TERMINAL-phase/grandchild identities, unrelated to the crash this PR fixes).

This PR's own fix (an uninitialized `prior_record_invalid` causing a `NameError` for the ephemeral-reviewer caller) was real and correctly diagnosed - it is NOT what caused the revert, and it does not by itself resolve reviewer-1's finding. But its target code no longer exists on master, so there's nothing left for it to patch. Leaving open rather than closing unilaterally in case the caller-audit in its description (exactly two callers of `_wrapped_liveness`, one of `evaluate_launch_barrier`, both confirmed by direct grep) is useful reference when the #163 redesign lands - happy to close on request.

zoolok17:
Closing: the title already says it — this hotfix was superseded when `627d5ee` reverted the change it was patching. Nothing here is lost; the underlying cold-start work is tracked as the "complete ownership proof" item in the v0.83.0 roadmap, which supersedes this approach rather than resuming it. Reopen if I have misread it.

PR #113, opened 2026-08-04
fix: capture wrapper logs on every launch path

### LINKED ISSUE #164: A mid-work status reply ends the wrapped turn and kills the running work: add an interim progress note that does not end the turn
## Observed (field, 2026-09-15 22:46Z)

A wrapped seat running an hour-long verification bar (two full test orders plus packaging) posted a status message mid-work ("mid-work, continuing now"). The reply ended the wrapped turn; the Maven run it had started died with the turn. An hour later the lead found zero test reports and no Java process. The seat was re-triggered and told never to send mid-work replies; the rule now lives in every dispatch.

## Expected

Either:
- an interim note the seat can post that the wrapper does NOT treat as the turn's reply (a progress kind, cursor untouched, turn continues), visible to the lead in `status` as "in progress since <t>, last note <text>", or
- a wrapper rule that a turn with an unfinished assignment continues after an interim reply, ending only on an explicit completion reply.

The first is simpler and also gives the lead the progress signal it currently lacks for long turns (the only signal today is the heartbeat).

Related: #162, #163; the heartbeat nudger workaround in the seat guidance.

zoolok17:
## Summary

- move bounded wrapper stdout/stderr capture into `agenttalk wrap` startup so direct/configured-argv launches are captured too
- retain one cross-language active-generation lock across the wrapper lifetime so Python and generated PowerShell pruning preserve live evidence
- record the root and generation that actually accepted capture, and expose it through `supervise --report --for <agent>`
- align direct and generated-launcher fallback order across preferred, OS-temp, and project-parent roots
- document the operator recovery path and the pre-dispatch boundary that remains tracked by #136

Closes #164.

## Regression evidence

Before the implementation, the new direct-launch and discovery integration tests both failed: the configured-argv wrapper produced zero log files, and the report had no recorded stderr location. With the implementation, a real non-supervised wrapper subprocess mirrors its console while producing committed bounded stdout/stderr files. The discovery test blocks the preferred root, captures in a fallback, makes the preferred root viable again, and still reports the fallback that actually received the generation.

## Targeted verification

- `python -m pytest tests/test_wrapper_logs.py -q`: 75 passed, 3 skipped
- `python -m pytest tests/test_supervisor.py -q -k "wrapper_log or ps_retention_preserves_generation_locked_by_direct_python_wrapper"`: 17 passed, 533 deselected
- selected supervisor init/report tests: 5 passed
- `python -m ruff check` on the five touched Python files: passed
- `git diff --check`: passed

## Residual

#136 does not fall out cleanly: wrapper-owned capture begins inside `cmd_wrap`, after Python import and argparse dispatch. Generated-supervisor OS redirection still captures earlier startup, but bootstrap/parse failures on a direct launch remain outside this change.

## Follow-up fix (commit c70e112)

CI on 7e3e04f was red on all 12 legs. Root-caused via the actual `dev-gate-evidence.json` artifacts, not just the console log:

- **`test_report_names_the_root_that_accepted_the_wrapper_generation` and `test_direct_allocator_refuses_a_symlinked_root_ancestor`** (12/12 legs): the CI evidence showed `targets.root` and `preferred.resolve()` landing on the *same safe fallback root* on every failing leg - no leak, refusal was already correct. The bug was that `default_wrapper_log_root()` delegated to the same reparse/symlink-filtered candidate search the allocator uses, so once refusal correctly fires and falls back, the "preferred" report and the real allocation always converge and the tests' `!=` assertion becomes unsatisfiable by construction. (Checked the prior commit, 461facf, too: its own CI evidence shows it genuinely did leak into the symlink target, so this was a real vulnerability that a later fix closed - the closing fix just broke this test's premise in the process.) Fixed by decoupling: `default_wrapper_log_root()` now returns the lazy, unfiltered first candidate (its only caller is artifact templating for the generated launcher, which independently refuses reparse points itself via `Get-SafeWrapperLogAgentDir` before ever using the baked value); `wrapper_log_root_candidates()` keeps its existing filtering for the allocator's real candidate list.
- **`test_direct_allocator_refuses_a_configured_state_junction`** (3 of 4 Windows legs, not 3.13): CI evidence shows a plain `subprocess.TimeoutExpired` - the test's own junction-creation PowerShell subprocess hit its fixed 20s budget under dev-gate load. Load-dependent, not a logic defect (matches the 3-of-4, non-3.13 pattern). Fixed by raising this and two related subprocess timeouts to 120s (matching this suite's existing PowerShell/subprocess-heavy precedent).
- `test_generated_ps1_survives_malformed_config_poll_with_last_good` (single windows/3.11 cell, source-run only, not wheel-run): left untouched - CI evidence supports flake, not a shared cause with the above. Confirmed gone on the next CI run.

CI re-run on c70e112 confirmed 3 of 5 fixed (both symlink/junction tests and the flake cell all green, Linux 4/4 and macOS 4/4 fully green). Two remained, both Windows-only, all four legs:

- `test_direct_wrap_launch_owns_bounded_log_capture` and `test_report_names_the_root_that_accepted_the_wrapper_generation` still failed. **The 120s timeout bump did not fix them** - this round's surviving traceback (these two sit where the evidence file's tail truncation doesn't eat them) showed `assert len(logs) == 1` / `assert 0 == 1`: the wrapper subprocess ran, exited 0, produced the expected stdout, and **zero log files existed anywhere under the test's tmp_path** - not a timeout, not a hang, and (checked explicitly) not a "right allocation, wrong reported root" divergence either, since the assertion does a global `rglob` across the whole tmp_path tree that would find the log wherever it landed.

## Investigation: zero logs found (commit 0d96ef7)

Leading hypothesis, evidenced but not CI-confirmed: **Windows `MAX_PATH` (260 chars)**, not a logic defect. CI's real failing `tmp_path` (`C:\Users\runneradmin\AppData\Local\Temp\agenttalk-dev-gate-<id>\pt-s-py310\test_report_names_the_root_tha0`) plus this code's own path contribution (a full 64-char sha256 `project_id`, plus a ~53-char generation-directory name) comes to 299 characters for the full `stderr.log` path - over the classic limit. Reproduced locally: an identical-length path succeeds via both `Path.mkdir()` and a raw `ctypes` `CreateDirectoryW` call with no `\\?\` prefix, but only because this machine has `HKLM\...\FileSystem\LongPathsEnabled=1` (checked, not assumed). If every candidate root shares the same over-length `tmp_path` prefix (true here - blocked-state, temp-fallback, and project-parent fallback are all rooted under the same tmp_path), all of them fail the same way and allocation reports total failure, matching "zero found anywhere" exactly. GitHub's actual runner configuration is not confirmed either way from public reports (mixed signal: some say long paths were enabled on their Windows images, others report continued failures on the same images) - genuinely a leading hypothesis, not a proven cause.

**Filing this as its own concern, not incidental debug output:** `_allocate_wrapper_log_targets` tried every candidate root and silently discarded each one's `OSError` before returning `None` with no reason. That collapsed "could not allocate, and here is why" into a bare "did not allocate," and is a real part of why this took multiple rounds and two refuted hypotheses to corner. Fixed permanently, independent of whether MAX_PATH turns out to be the cause: the allocator now raises `WrapperLogAllocationFailed`, carrying every candidate root and its actual `OSError` string; `WrapperLogInstallation` gained a `disabled_reason` field carrying that detail through `installed_standard_streams_from_environment`; `cmd_wrap` prints it to the (still-unwrapped, in this failure path) real stderr so it is visible to any operator hitting this, not just to CI. The two affected tests' assertion messages now include the wrapper's stderr too, so the next CI run either confirms MAX_PATH with the actual `WinError` in hand or hands us the real reason instead of another hypothesis - no fix has been chosen yet pending that evidence.

## MAX_PATH-as-total-failure: REFUTED by the diagnostic above (commit ec4e3f6)

CI re-ran on 0d96ef7. `test_report_names_the_root_that_accepted_the_wrapper_generation`'s traceback survived this time (windows/3.10, both source and wheel): `result.returncode == 0` passed, the wrapper's stderr was a single valid lifecycle JSON line (`"event":"wrapper_exited","exit_code":0,...`), and **there was no `disabled_reason` text anywhere in it.** The diagnostic added above proved itself by absence: it is verified to print the reason on every total-allocation-failure path, and the reason is not there. Conclusion, not another guess: **allocation did not totally fail. `_allocate_wrapper_log_targets` returned targets.** MAX_PATH-as-the-cause-of-total-failure is refuted, on this run, by the instrument built to test it. (The `pt-s-`/`pt-w-` short-basetemp-naming corroboration cited above pointed the same wrong way and is retracted as support for this specific hypothesis.)

That still leaves the actual symptom unexplained: if allocation succeeded, a log was written *somewhere*, and `_captured_stderr_logs(tmp_path)` (a global `rglob` over the whole tmp_path tree) still found zero. Three live explanations, none chosen:
- **(A)** The accepted candidate root is outside `tmp_path` because the test's Windows env isn't fully self-contained the way it is on POSIX - would also explain the platform split without needing a path-length story at all.
- **(B)** The preferred root was rejected (possibly still MAX_PATH-related, possibly not) and a fallback candidate *outside* `tmp_path` was accepted instead - keeps MAX_PATH alive as a contributing cause while still refuting it as the reason nothing was found.
- **(C)** The log landed inside `tmp_path`, just under a name/layout `_captured_stderr_logs`'s glob doesn't match.

All three share one defect: the diagnostic above only reports *total* failure. A rejected-preferred-root-then-quietly-accepted-fallback is exactly as silent as total failure was before `disabled_reason` existed - the same class of problem one level down. Commit ec4e3f6 closes that gap: `WrapperLogInstallation` now also carries the accepted root and every rejected-then-superseded candidate (with its real `OSError`) whenever the direct-launch allocator actually ran, and `cmd_wrap` prints both to stderr. Verified directly against a real subprocess run with a blocked preferred candidate before pushing: stderr shows both the rejected candidate's `WinError 183` and the accepted fallback's full path. The next CI run's `result.stderr` (already embedded in both tests' assertion messages) says either "accepted `<path outside tmp_path>`" (A or B - confinement is the bug) or "accepted `<path inside tmp_path>`" (C - the test's own collector is the bug). Still no fix shape chosen pending that answer.

## Closed: (A) confirmed, fixed by reading the fact instead of guessing the directory (commit 791f4f1)

CI re-ran on ec4e3f6 and answered all three hypotheses at once. The rejected candidate's error is `[WinError 183] Cannot create a file when that file already exists: '...\blocked-state'` - the test's own deliberately-blocked file doing its job, not a length error; MAX_PATH is dead a second way. The accepted root is `<dev-gate basetemp>\agenttalk-wrapper-logs\<project-hash>` - a sibling of the test's own `pt-s-py310`/`pt-w-py310` basetemp, one level *above* `tmp_path`. That is **(A)**: the capture path did exactly the right thing (preferred root blocked, fell back, succeeded) on every one of the red Windows legs. There was never a capture failure - only a test that located the evidence by guessing a directory instead of asking the wrapper what it actually used.

Fixed by reading the fact: both tests now call `wrapper_logs.read_wrapper_log_location()` (the same persisted record `supervise --report` reads, and the same pattern `test_direct_wrap_captures_implicit_root_discovery_failure` already used) instead of `_captured_stderr_logs(tmp_path)`'s `rglob`. `test_report_names_the_root_that_accepted_the_wrapper_generation` - literally named for the root the wrapper says it accepted - now asserts against that named fact rather than an inferred directory, which is the property the test was always meant to check and stays true no matter where a legitimate fallback resolves. Verified this actually distinguishes the fixed case from the broken one, not just that it compiles: reproduced the CI shape locally by also blocking the temp-fallback candidate, forcing `tempfile`'s own internal fallback to a real system location (`C:\Windows\Temp\...`) outside `tmp_path` entirely, and confirmed the new assertions pass against that location precisely because they read the recorded fact instead of a guessed directory.

Applied the identical fix to `test_direct_wrap_launch_owns_bounded_log_capture` on the strength of shared reasoning, not direct confirmation: its own CI traceback is still eaten by evidence truncation (it is not the last-defined failing test in the file, same limitation as every prior round), so I could not confirm its failure was the identical mechanism this round. The fix is correct regardless of the exact cause - reading the recorded fact is right whether or not this specific test hits the same escape - and if it shares the cause, this closes it too; the next CI run will show directly whether it's now green.

**Hygiene, not chosen as the fix:** a real subprocess writing outside its own `tmp_path` on a CI runner is worth a second look independently of this bug - it leaves litter and could theoretically collide with a concurrent leg (mitigated in practice: the log root is project-hashed, and production pruning already bounds the generation count per root). Did not confine the test environment to force fallback candidates to stay under `tmp_path`, on the view that doing so would make the test pass by narrowing the scenario rather than by checking the right thing, and would stop exercising the real cross-boundary fallback path this bug lived in. Noted, not acted on.

**Opinion on whether a shared-temp fallback is the right production behavior (not acted on in this branch, as instructed):** the fallback shape itself is defensible - it's project-hashed (no collision), and *something* to fall back to beats the silent total-failure this investigation spent most of its rounds ruling out. The real gap was discoverability, not the fallback existing: before this PR's diagnostic commits, an operator whose preferred root got silently rejected had no way to learn a fallback even happened, let alone where. `disabled_reason`/`rejected_attempts`/the accepted-root stderr line already close most of that gap for anyone reading wrapper stderr or `supervise --report`, which already existed for exactly this purpose. I would not prioritize moving the fallback out of the OS temp directory over making sure operators know to look; happy to be argued out of that if the temp directory's own volatility (cleanup tools, reboots on some setups) is a bigger concern than I'm weighing it as.

## Cold review, first-ever pass on this PR (commit 9b8435e)

This PR reached 15/15 green with zero prior review. A cold pass came back with three P2s, none of which are on `master` yet (the whole feature is unmerged) but all of which predate my four commits - blamed to dev-8 (`git blame`: item 1 and item 2 to `7e3e04f`, item 3 to `461facf`, both before I took this PR over). Confirmed each against the code (a real repro or authoritative source-reading, not agreement-by-default), then split by disposal:

- **Implicit-launch retention pool keyed by cwd, not the discovered project root** (`cli.py` implicit-root handling): confirmed the mechanism - `agenttalk wrap` launched without `--root` hashes the provisional `Path(".")` into the allocation's project_id before `find_root()` ever runs; the location record gets moved to the discovered root's `.agenttalk/state/` afterward, but the record's own `root`/`generation_dir` fields still point at the cwd-derived, disjoint pool. **Split into its own task (#172), not fixed here.** The straightforward-looking fix (resolve the project root before choosing the pool) cuts directly against a comment already in this code - "Do not perform the authoritative, fallible root resolution before bounded streams exist" - a deliberate fail-soft ordering decision, not an oversight. Reconciling "capture must work even if discovery later fails" with "capture must count against the right project's quota" is a design call, not a patch to slip into an otherwise-finished PR.

- **`default_wrapper_log_root`/`wrapper_log_root_candidates` raise/return-empty for a project rooted at a filesystem anchor** (no same-volume "outside" exists, so every candidate resolves as relative to the project): **fixed.** `git diff 461facf 7e3e04f` shows the exact block dev-8 removed in the same commit that added the raise this replaces - no rationale given, reads as incidental during the reparse-filtering rework. Restored the same unconditional last-resort fallback in both functions. This is restoring a guarantee that existed and was silently dropped, not inventing one - said here plainly rather than just fixing it quietly, since my own earlier rewrite of `default_wrapper_log_root` (commit c70e112) preserved the raise without reintroducing the fallback, because I did not know it had been there until this review surfaced it.

- **`_owned_committed_generations` treats an inaccessible `agent_dir` as confirmed-absent, not unknown**: **fixed properly**, per the review's own suggested direction and the same family as this round's earlier diagnostics. `Path.exists()` is a boolean wrapper that swallows some `OSError`s into `False` - confirmed directly against CPython's own `pathlib` source rather than assumed: only a curated list on 3.10 (`ENOENT`/`ENOTDIR`/`EBADF`/`ELOOP` plus `WinError 21` "not ready" on Windows - a disconnected network share or removed media), but a bare `except (OSError, ValueError)` on 3.14+, where `Path.exists()` delegates to `os.path.exists()`. Consequence: `max_sequence` under-counts, the next generation's sequence number comes out too low, and when the transiently-unreadable root recovers, its real higher-sequence generations can outrank and *prune* the genuinely-newer one - data loss in the retention path. Fixed by probing with `lstat()` directly (the raw, non-swallowing primitive) instead of `exists()`/`is_dir()`: only `FileNotFoundError` means genuinely absent, matching the sibling non-directory/reparse branch two lines down that already treats everything else as uncertain.
  - **Closed the dependency this fix would otherwise leave inert** (review comment #5, confirmed rather than just noted): `grep` across the whole codebase found exactly one reference to `.sequence-uncertain` - the write site itself. Nothing read it back. `_read_wrapper_log_sequence` (and its PowerShell mirror, `Read-WrapperLogSequenceRecord`) now OR the marker into every return path's uncertainty, so a generation marked uncertain at allocation time stays uncertain on every later scan regardless of whether `.sequence` itself later reads back as well-formed. Without this, the `_owned_committed_generations` fix above would have recorded the uncertainty correctly and changed no retention decision - a check that runs but is never read is not a check.

Verification note on my own method: my first repro attempt for the inaccessible-agent-dir fix used a plain `PermissionError`, which does **not** actually reproduce the bug on Python 3.10 (`EACCES` is not in `pathlib`'s ignored-error list, only `WinError 21` is) - it passed on both the pre-fix and post-fix code, which would have been a false-positive test had I not checked. Rebuilt the repro with the exact exception class `exists()` actually swallows, then verified every new/changed assertion failing-first against the pre-fix source (via a scoped `git stash` of just the source file, tests held constant) before accepting it as passing against the fix.

Verification: `tests/test_wrapper_logs.py` (89 passed, 3 skipped - up from 82, all new coverage) and the wrapper-log/sequence/retention-tagged slice of `tests/test_supervisor.py` (58 passed, including one new PowerShell-side test mirroring the Python-side fix) on Python 3.10 and 3.14; ruff clean.


zoolok17:
@codex review

zoolok17:
@codex review

zoolok17:
@codex review

zoolok17:
@codex review

zoolok17:
@codex review

zoolok17:
@codex review

zoolok17:
@codex review

zoolok17:
@codex review

zoolok17:
@codex review

zoolok17:
Closing the 6,210-line draft (recorded status check 2026-09-17, operator-approved triage) and keeping the logging guarantee as a narrow issue: https://github.com/zoolok17/agenttalk/issues/181. Branch kept.

PR #116, opened 2026-08-04
fix(#166): refuse loudly with a remedy that is provably right — or say plainly what is not verified

zoolok17:
When the supervisor refuses to relaunch a wrapped agent, it used to say nothing useful. This makes it print one bounded detached-launch instruction when available, or say why no scripted remedy applies. The two recovery populations have different contracts: active ephemeral recovery verifies prepared evidence; regular process-tree-HOLD recovery reconstructs current static configuration without a prepared request or effective binding.

## What active ephemeral recovery IS bound to

Before emitting an active ephemeral recovery command, agenttalk verifies all of the following:

- the **prepared request** (immutable request evidence and the durable one-shot review message)
- the **marker-derived substitutions**, read positively from the live marker — the archived original is used only after absence is *proven*, not merely unreadable
- the **configured launch mapping**, still raw-equal to the prepared mapping, and the **effective CLI selection by resolved executable path**, still equal to the prepared selection
- the **reconstructed launch-effective projection**, including fully substituted argv and the emitted working-directory value

**Re-prepare after editing the configured profile.** Active ephemeral recovery emits a named refusal rather than a command when the configured launch mapping or reconstructed launch-effective projection no longer matches; equivalent-looking edits are not guaranteed to remain valid.

Active ephemeral recovery re-runs full admission, reconstructs the same projection, and requires exact equality. Missing, malformed, superseded or unknown-version bindings **fail closed** — no command is emitted, and a named reason is given instead. Nothing is backfilled; absent evidence is never invented.

The emitted working directory is always absolute. Without a lane, it is the admitted profile or project-root value. With a lane, it is the bound lane workspace and may differ from the configured profile value.

## The regular process-tree-HOLD recovery contract

Regular recovery validates and reconstructs the current static configuration, but has no prepared request or effective binding to compare. A relative configured working directory is emitted unchanged; run the command from the same base the supervisor used, or the agent can start in the wrong directory. An absolute working directory has no such base hazard, but regular recovery does not check that it exists.

## What recovery is explicitly NOT bound to

**Executable contents are NOT verified. Active ephemeral recovery binds resolved executable paths, not the contents at those paths.** An in-place replacement at one of those paths is not detected, so a recovery command may name a binary whose contents have changed since preparation.

**The environment the child process actually receives is NOT verified.** Review it yourself before running the command.

For active ephemeral recovery, a matching launch-effective projection is **not** evidence that ambient or configured values were delivered to the child. Any environment shown for either population is guidance, not a guarantee.

This is deliberate. Three attempts to make that environment guarantee true each failed a different way — too strict (a single non-English variable name would have disabled every launch), then removed while the promise stayed, then still untrue at the real child boundary. Two platform facts killed it: the preferred PowerShell silently drops hidden per-drive entries across the process boundary, and Windows PowerShell 5.1 cannot represent an admitted empty value through its provider. Both were shown with native child probes. The claim is narrowed rather than chased a fourth time; the work is tracked separately.

## Also fixed

- **A partial failure while applying the environment used to leave the parent process mutated.** Restoration is now armed before the first change, the spawn is suppressed on failure, and absent / empty / non-empty values are each restored exactly — via native calls, because the older host cannot round-trip an empty value.
- **A destructive command trusted a state file the caller supplied**, made real changes, then recorded them only in the caller's copy. Both the create and archive paths now refuse any non-canonical state path before reading. **Archive additionally** holds the lifecycle lock and verifies exact live instance identity; create does neither — that gap is real and tracked, not claimed as closed.
- Cross-platform projection: Windows paths are parsed as Windows paths rather than resolved against whichever machine is describing them.
- A malformed config exits with the documented input-error code again, not the hold code.

## Known and deliberately not fixed here

A failure at the final save of a prepare can still leave real effects with no canonical record. That is pre-existing, needs a durable journal rather than another lock, and is tracked. **This boundary is not described as atomic or closed anywhere, and must not be.**

## The rule this release ended up being about

The text must never be stronger than the enforcement. Most of what was found reviewing it — including by its own author and lead — was a promise outliving the thing that made it true.


zoolok17:
@codex review

zoolok17:
@codex review

zoolok17:
@codex review

zoolok17:
@codex review

zoolok17:
@codex review

zoolok17:
@codex review

zoolok17:
@codex review

zoolok17:
@codex review

zoolok17:
**Correction to 7c2cd48's commit message, from the courier.**

That commit's message describes only the cross-platform projection fix and the template-expectation change. It also contains — undescribed — the `--prepare-launch-request` canonical-state boundary fix in `cli.py`: rejecting a non-canonical `--state-file` before any read, and loading/saving the derived canonical path rather than the caller-supplied string.

Cause: I committed the author's worktree while they were mid-turn on a follow-up instruction, and my message was written from their earlier report. The code is the author's and appears coherent; the omission is mine.

Note for anyone reading this PR later: **7c2cd48 carries that production change WITHOUT its regression control.** The before-effects copied-state test is required and lands in a following commit.

zoolok17:
**Correcting my own correction, from the courier.**

My earlier comment on 7c2cd48 said that commit carried the `--prepare-launch-request` boundary change **without** its regression control. That is wrong.

The author verified against the commit rather than the working tree: `test_prepare_cli_rejects_copied_state_before_effects` was introduced **by** 7c2cd48 and is absent from its parent. It drives the real CLI with a copied state path and asserts the official state, copied state, live marker, archive, project config, retired roster and message set are all unchanged. Production change and control landed together.

So the accurate record for 7c2cd48 is: its message omits the prepare-boundary change, but the change is complete and controlled. I overstated the harm in the direction of alarm, which is its own kind of inaccurate. Both corrections stand together.

zoolok17:
@codex review

zoolok17:
@codex review

zoolok17:
@codex review

zoolok17:
@codex review

zoolok17:
@codex review

zoolok17:
@codex review

zoolok17:
@codex review

zoolok17:
@codex review

zoolok17:
@codex review

zoolok17:
@codex review

zoolok17:
@codex review

PR #118, opened 2026-08-06
feat(#169): let a wrapped codex agent commit/push its own worktree

zoolok17:
## Summary

#169: a wrapped codex agent can now `git add`/`commit`/`push` from its own git worktree, under its own identity, without the operator's approval-escalation workaround.

## The mechanism (measured, not re-derived)

Agent worktrees live outside the main checkout (e.g. `D:\tmp\at-task166-loud-refusal`). A linked worktree's git metadata does not live in the worktree - it lives in the MAIN repo at `<main>\.git\worktrees\<name>\`, outside the agent's granted `writable_roots`. Under codex `sandbox_mode = "workspace-write"`, that write is denied: `fatal: Unable to create '.../.git/worktrees/<name>/index.lock': Permission denied`. Not a filesystem-ACL problem (already ruled out) - it's the sandbox's own write-side policy, keyed on `writable_roots`.

## Design

Extended the existing seeding mechanism (`supervisor.codex_config_overlay`, called by `agenttalk supervise --seed-codex-config`, which already writes `writable_roots` into the per-agent isolated `CODEX_HOME/config.toml`) rather than inventing a parallel one. A new pure helper, `codex_worktree_writable_roots(repo_path)`, computes what to add:

- If `repo_path` is ITSELF a linked worktree, grants exactly its own admin dir (`<main>/.git/worktrees/<this-one>`) - narrow, no other worktree's metadata is touched.
- Otherwise (the case every currently-seeded agent is actually in: `repo_path` is the shared team root, and agents create worktrees elsewhere, ad hoc, after seeding, under names not known yet), grants the whole `<repo_path>/.git/worktrees/` parent. **This is wider than one agent's own worktree** - it also exposes every OTHER agent's worktree-scoped git state (index, HEAD, per-worktree logs) under the same repo. There is no narrower target that survives every worktree an agent might create later without inventing a second, worktree-creation-triggered reseed path alongside the existing launch-time one - deliberately not built, per "start from the existing mechanism."
- Either way, ALSO grants `.git/objects`, `.git/refs`, `.git/logs` - git's shared, cross-worktree object/ref/reflog storage that any worktree's commit/push writes into by git's own design (worktrees share history on purpose). **This is the part that can't be narrowed further at all**: it grants filesystem-level write access to any ref or object in the repository, not only the seeded agent's own branch. Git offers no filesystem-level way to scope tighter than "the shared object/ref store."

None of this ever touches `.git/config`, `.git/hooks`, `.git/info`, or anything outside `.git` - and never touches the operator's actual source tree beyond what `writable_roots=[<repo_path>]` already granted before this PR (asserted directly in tests, not just reasoned about).

`codex_config_overlay` gained `extra_writable_roots: tuple[str, ...] = ()`, rendering additional `writable_roots` array entries. Default-empty, so every existing caller/test is byte-identical. Idempotent (re-applying the same extras is a no-op; re-seeding with different extras replaces the array rather than accumulating it) - survives `--refresh-scripts`/reseed the same way the existing managed keys do, because it's the SAME seed step, not a new one.

## Verification

**Positive, failing-first, and via TWO independent methods:**
- Unit/CLI-level (pytest): `codex_worktree_writable_roots` and `codex_config_overlay` against a hand-built linked-worktree `.git` layout (no git binary needed - these are pure file readers), plus a full `agenttalk supervise --seed-codex-config` CLI run against a REAL `git worktree add`-created worktree, asserting the rendered `config.toml` contains all four extra roots.
- Filesystem-level, against a REAL git repo with a REAL linked worktree: computed `codex_worktree_writable_roots()`'s actual output, applied those EXACT paths as Windows ACL write-grants (denying write everywhere else in `.git`, using narrow WriteData/AppendData/DeleteChild rights rather than the broader `W` alias, which also blocks reads codex's sandbox doesn't restrict), and confirmed `git add && git commit && git push` succeeds end-to-end from the worktree. Failing-first: reproduced the exact reported `index.lock: Permission denied` failure with the deny in place and no grant, before granting the computed roots and watching it turn to `COMMIT OK` / `PUSH OK`.

**Negative, tested for real, not just reasoned about:**
- A SECOND worktree's own admin dir (`.git/worktrees/<other-name>/`) stays denied under the exact same grant set - confirmed with a real `index.lock: Permission denied` on the sibling, in the same test session as the positive control's success.
- `codex_worktree_writable_roots()`'s output is asserted (unit test, not just eyeballed) to never include the checkout's own working-tree root - every returned path is `.git`-relative.

**What I could not verify, said plainly rather than glossed over:** I could not run the fix against codex's actual Windows-restricted-token sandbox end-to-end - I have no way to execute AS a codex-sandboxed process myself. I tried `codex sandbox -P workspace-write -C <dir> -- git ...` (a real debug subcommand shipped with the installed codex CLI, 0.144.6) to test directly, but it requires a `[permissions]` table under a newer named-profile schema I could not get working from the legacy `sandbox_mode`/`sandbox_workspace_write.writable_roots` config this codebase seeds - possibly a newer, parallel permission system in current codex releases that I don't have documentation for. The Windows-ACL-based verification above is a reasonable proxy for a write-side-only restriction (deny writes outside listed roots, never restrict reads) rather than the real mechanism, and it produces the exact reported failure signature and the exact reported fix signature, but it is a proxy. Recommend one live check with an actual wrapped codex agent before treating this as fully closed.

## What has to change on the operator's machine

Nothing in configuration. This is baked into the existing `Seed-CodexHome`/`--seed-codex-config` step, which the generated supervisor PS1 already re-runs on every launch AND every relaunch (crash-recovery, idle-timeout restart, manual restart) - so any codex agent that restarts after this PR is deployed picks up the fix automatically, no new trigger needed.

For agents that are CURRENTLY RUNNING and won't naturally restart soon: their CODEX_HOME's config.toml was seeded under the old code and stays stale until reseeded. Two ways to get them current without waiting:
- Restart the agent (supervisor's normal relaunch path reseeds automatically), or
- Manually run `agenttalk supervise --seed-codex-config --home <that agent's CODEX_HOME> --repo <its --root>` against the upgraded agenttalk code.

## Not acted on (per scope note - this is an enabler, not a release blocker)

None of the above touches sandbox_mode, approval_policy, or anything for Claude-family agents.

### DIFF
--- a/src/agenttalk/cli.py
+++ b/src/agenttalk/cli.py
@@ -11985,8 +11985,14 @@ def cmd_supervise(args: argparse.Namespace) -> int:
             cxc.repair_duplicate_project_tables(cfg_p, Path(repo))
         existing = cfg_p.read_text(encoding="utf-8-sig") if cfg_p.exists() else ""
         cfg_p.parent.mkdir(parents=True, exist_ok=True)
+        # #169: also grant whatever git-worktree admin state this agent needs
+        # to commit/push its own worktree - see codex_worktree_writable_roots
+        # for exactly what that is and isn't.
+        extra_roots = sup.codex_worktree_writable_roots(repo)
         cfg_p.write_text(sup.codex_config_overlay(existing, repo_path=repo,
-                                                  windows_sandbox=sandbox), encoding="utf-8")
+                                                  windows_sandbox=sandbox,
+                                                  extra_writable_roots=extra_roots),
+                         encoding="utf-8")
         print(f"seeded codex config.toml (sandbox={sandbox}): {cfg_p}")
         return 0
     if args.seed_claude_settings:
--- a/src/agenttalk/supervisor.py
+++ b/src/agenttalk/supervisor.py
@@ -12443,16 +12443,144 @@ def _toml_basic_str(value: str) -> str:
     return '"' + value.replace("\\", "\\\\").replace('"', '\\"') + '"'
 
 
+def _linked_worktree_admin_dir(path: Path) -> Path | None:
+    """Return the admin dir a linked worktree's ``<path>/.git`` points at.
+
+    A linked worktree's own ``.git`` is a plain FILE containing
+    ``gitdir: <admin-dir>`` (git's own mechanism for finding its per-worktree
+    metadata, which lives elsewhere - see ``git worktree`` documentation).
+    Returns ``None`` (fail-soft, never raises) if ``<path>/.git`` doesn't
+    exist, is a directory (an ordinary checkout, not a linked worktree), or
+    can't be read/parsed - callers treat that as "nothing extra to grant",
+    not as an error.
+    """
+    git_marker = path / ".git"
+    try:
+        if not git_marker.is_file():
+            return None
+        text = git_marker.read_text(encoding="utf-8").strip()
+    except (OSError, UnicodeError):
+        return None
+    prefix = "gitdir:"
+    if not text.startswith(prefix):
+        return None
+    raw = text[len(prefix):].strip()
+    if not raw:
+        return None
+    try:
+        return (path / raw).resolve() if not Path(raw).is_absolute() else Path(raw).resolve()
+    except (OSError, RuntimeError):
+        return None
+
+
+def _worktree_common_dir(admin_dir: Path) -> Path | None:
+    """Resolve the shared ``.git`` common dir a worktree admin dir points at.
+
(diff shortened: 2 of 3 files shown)


zoolok17:
Closing on operator approval of the 2026-09-17 backlog triage: superseded in practice - codex seats commit and push their own branches today (PRs #176 and the #170 follow-up were produced that way on v0.88.0) without this change. Reopen from a concrete failure.

PR #119, opened 2026-08-07
feat(#171): keep the part of a failure that says which test failed

zoolok17:
When a gate leg goes red, the CI log says only `exit code 1`. The stored evidence was meant to carry the rest, and it kept the **last** two thousand characters of output — which on a real failure is the tail of a diff.

Measured on a real 12,718-character failure: the retained text contained none of the test name, the assertion, the failing frame, the FAILURES header, or the FAILED token.

**The bound was not the defect. Keeping the wrong end was.** The two-thousand-character ceiling is unchanged on purpose — raising it would have hidden the real problem behind more text.

## What changed

- A gate-owned reporter captures the first failure the moment pytest reports it. The record is now a labelled envelope in priority order: which test, which phase, which frame, the assertion, how many further failures were omitted, an excerpt — and only then whatever tail still fits.
- A second, independent loss is closed: quiet mode elides comparison detail *before* our code sees it, so the selector was choosing from data pytest had already discarded. Assertion verbosity is now on.
- The aggregate no longer replaces a leg's diagnosis with "CI leg blocked".

## Fail-safe

This machinery lives inside the gate, so it must never be the reason a gate fails. A missing, empty, corrupt, oversized, incomplete or killed capture falls back to the previous bounded tail **and labels the record as having done so** — a diagnostic that fails silently would make an unhelpful artifact look like a deliberate one.

## Deferred deliberately

Publishing the same summary as build annotations (which outlive artifacts by a wide margin) adds an output channel with a text-injection surface, and gets its own review rather than riding along here.

## Evidence

Failing-first: the new selector regressions failed before the capture existed. 114 gate tests pass. Two independent reviews approved; both found real issues that were fixed first — a lost right-hand assertion value, exception objects whose string form could alter pytest's verdict, oversized sidecars fully allocated before rejection, and lone-surrogate data that could break JSON serialisation.

Authored by codex-agenttalk-developer-8, couriered by the operator.

### DIFF
--- a/dev-gate.json
+++ b/dev-gate.json
@@ -67,7 +67,9 @@
         "tests"
       ],
       "args": [
-        "-q"
+        "-q",
+        "-o",
+        "verbosity_assertions=2"
       ],
       "test_requirement": "pytest>=8.0",
       "timeout_seconds": 2700
--- a/docs/DEV-GATE.md
+++ b/docs/DEV-GATE.md
@@ -88,14 +88,20 @@ the evidence never presents them as content-addressed inputs.
 ## Evidence contract
 
 Run artifacts use `artifact_type: agenttalk-dev-gate-run`. They contain exact required check IDs, commands,
-resolved tool paths and versions, exit codes, durations, log hashes and tails, isolated-venv creator/prefix
-proofs, pip-configuration and child-`PATH` isolation assertions, import provenance, package
+resolved tool paths and versions, exit codes, durations, log hashes and bounded diagnostics, isolated-venv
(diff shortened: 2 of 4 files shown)


zoolok17:
@codex review

zoolok17:
Closing on operator approval of the 2026-09-17 backlog triage: superseded by PR #176 (issue #170), which uploads the complete per-check logs and re-verifies them at read time; this draft kept only a bounded first-failure summary and never uploaded the log. The first-failure summary idea is recorded on #170 as an optional follow-up.

PR #120, opened 2026-08-07
feat(#128): keep the bytes the gate actually tested

zoolok17:
Every gate leg builds a wheel and an sdist, records their hashes, and **throws the bytes away**. So no release this project has made can prove it published what was tested — the packages were rebuilt afterwards, on a laptop, from a tree nobody re-checked. The previous release has no durable evidence attached at all.

The same measurement found a second thing: **at the last tag, the install instructions still named a version two releases old.** Corrected afterwards. Not a hypothetical failure of a hand-run ritual — one that already happened, in the first thing a user reads.

## Deliberately the read-only half

Re-gates one named commit — full 40-character SHA, refused if it isn't the current master tip, isn't the checked-out head, or is a partial rerun — and keeps **the exact wheel and sdist the gate built**, with all twelve leg records, the aggregate, a same-run receipt and checksums. It does not rebuild the packages. It does not tag, publish, or write anything to the repository.

Tagging stays manual until this has produced a record somebody has read. Trusting an automated release path before its evidence has ever been examined would repeat the mistake it exists to prevent.

## Three refusals, kept apart

Evidence missing, evidence stale, and evidence bound to a different commit are three distinct refusals. Collapsing them is this project's signature defect, and a release path that couldn't tell an operator which had happened would be the worst possible place for it.

## Note on authority

With a single-person roster, a required-reviewer gate is the operator approving their own dispatch — it is a deliberate second action, not two-party control, and the design says so rather than dressing it up.

Authored by codex-agenttalk-developer-8, couriered by the operator. The second commit adds two test files the first one missed — my error, recorded there.


zoolok17:
Closing on operator approval of the 2026-09-17 backlog triage: stale August draft. The idea is preserved as a fresh issue (release provenance: keep or hash-join the tested wheel and sdist), ranked after the blind rerun. Branch kept for reference.

PR #121, opened 2026-08-08
fix(#181): require live owner proof before automatic archive retires an agent

zoolok17:
## What this fixes

Automatic archive of an ephemeral launch request could retire a **live** agent. It validated the request id, the state file, the terminal label and optional completion evidence — then archived, unlinked the active launch marker, removed active supervisor state and retired the roster identity, **without ever checking that a live owner existed**.

Reproduced against a clean export of `master` with a real live child PID in official supervisor state: archive returned `rc=0`, the claimed process was **still running**, the marker was gone, active state was gone, the identity was retired, and a subsequent send from it failed as a tombstone.

## Severity, stated honestly

- **Current v0.81+ normal operation does not fire this.** The owned-tree evidence gate holds the request active on missing, corrupt, ambiguous or surviving-owner evidence, which defeats the completion race, timeout, crash/restart, duplicate-poll and second-supervisor cases.
- **The current trigger is a direct or replayed internal command** (`supervise --archive-launch-request …`) while the reviewer is live. Plausible during manual troubleshooting or local automation; not a documented ritual.
- **v0.36–v0.80 had a genuinely automatic trigger**: a transient process-table capture failure yielded no kill targets, so the executor skipped the stop and called raw archive with no barrier. Closed by `587e7c1`, first shipped in v0.81.0.

No repository artifact shows an actual field occurrence, so this is **not** a claim that supervisors are spontaneously retiring live agents today.

## What the operator lost when it fired

- An **unmanaged orphan**: the process is not killed and can finish a paid turn while invisible to status, report and attention.
- **Destroyed work** if archive won the race against publication — the store rejects a retired sender, so the typed result and the owed reply are lost.
- The identity cannot be recovered or rebound; a later request gets a different one.
- Lane idleness is derived from the active launch marker and supervisor state. Erasing both can declare a **live** lane-backed reviewer idle, so a later remove can run beneath it.

## The change

One **positive** five-outcome owner probe — `alive`, `dead`, `pid_reused`, `unknown`, `start_unmatchable`. It never inverts `owner_gone`, which deliberately collapses live and unknown so teardown fails safe; reusing it would have reproduced this defect wearing a guard's name. An absent start is `start_unmatchable`: PID liveness alone cannot prove the recorded process identity.

Automatic archive now requires exact identity equality with the strict instance marker, holds the lifecycle lock, positively probes the recorded owner, **strictly rereads the marker to close the probe/swap race**, and only then loads state or performs any effect. Dead, reused, unknown and unmatchable each refuse with a distinct stable code. All four generated PowerShell archive calls carry the claimed identity.

**Attended archive is untouched** — it has the opposite authority contract (supervisor absent, operator-attested) and keeps its passing control.

## The shipped test that certified the bug

`test_cli_archive_launch_request_preserves_completion_evidence` called this with **no identity at all** and asserted `rc=0`, removal and retirement. It has passed every review since 2026-06-27 while pinning the defect as correct. Replacing it is the point of this change, not a side effect: it is now a no-effect refusal plus a live-positive success.

The later `test_cli_archive_launch_request_allows_live_owner_without_start_token` then certified the same defect one level deeper by treating a live numeric PID without a recorded start as owner proof. It is now a full-byte null-start refusal control, alongside a positive public drain control proving the supported generic null-start claim/release path remains intact.

Added byte-for-byte no-effect controls for the launch marker, instance marker, archive tree, config, active roster, retirement registry and supervisor state; plus stale PID, PID reuse, unknown, unmatchable, marker swap, null-start refusal, public null-start drain compatibility, **owed-result publication** and **lane-busy preservation**.

## Residual risk

- This proves the **singleton supervisor owner** is live; it does not independently prove the ephemeral worker is dead. A deliberate replay carrying a genuinely live supervisor's identity still passes. Worker safety remains the kill/teardown barrier.
- The generated supervisor claim is Windows-only today. The probe works with native Linux start tokens and fails closed when no comparable start token exists, but a future POSIX supervisor must canonicalize its start token and needs a real platform test.
Automatic archive still retires the roster identity before publishing and unlinking the archive, and before the later supervisor-state save. That ordering and its partial-effect failure window are pre-existing at base `a8f5f468`, remain unfixed here, and belong to the deferred journal work. This change guarantees byte-identical no-effect behavior only for the new pre-admission refusal paths; a strongly admitted archive can still return `rc=3` after partial effects if archive publication, unlink, or the final state save fails.
- `ruff format --check` wants to reformat the touched baseline files wholesale; that churn was deliberately not introduced here.

Scope: production +138/−18, controls +503/−30.

Authored by codex-agenttalk-developer-8; committed in its own clone and pushed by the operator as courier.

### DIFF
--- a/src/agenttalk/cli.py
+++ b/src/agenttalk/cli.py
@@ -26,6 +26,7 @@ from pathlib import Path
 
 from agenttalk import __version__
 from agenttalk import avatars as avatar_mod
+from agenttalk import store as store_mod
 from agenttalk.display import render
 from agenttalk.store import (
     COMPOSING_INTENT_STALE_SECONDS,
@@ -12472,6 +12473,13 @@ def cmd_supervise(args: argparse.Namespace) -> int:
                              "--request-id <rid>, --terminal-state <state>, and "
                              "--state-file <path>\n")
             return 2
+        if not args.instance_token or args.pid is None:
+            sys.stderr.write(
+                "agenttalk supervise --archive-launch-request: "
+                "supervisor_owner_identity_missing: need the live supervisor "
+                "--instance-token and --pid identity\n"
+            )
+            return 3
         completion = None
         if args.completion_json:
             try:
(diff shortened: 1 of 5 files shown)


zoolok17:
@codex review

zoolok17:
@codex review

PR #130, opened 2026-08-27
fix(#184): status reads the message store once, not 24 times

zoolok17:
Replaces the per-agent unread_for loop (one near-full store scan per roster agent) and four further independent full scans with ONE shared validated snapshot threaded through backward-compatible optional parameters.

Measured on the live 7,038-message store: ~8-9s -> ~1.4s (the reported 46s worst case scales with store size x stale-roster size).

Evidence: byte-identical `status --json` proven by running the same fixture test against pre-fix code; a scan-count spy asserting exactly one snapshot per invocation (which caught a fifth hidden scan pre-review); 486-540 targeted tests green untouched.

Review: approved by claude-agenttalk-reviewer-3 (rq-d2512b82c526) with ID-list equivalence probes, a disk-walk counting probe at a lower primitive, and full caller enumeration of the five signature changes.

### DIFF
--- a/src/agenttalk/cli.py
+++ b/src/agenttalk/cli.py
@@ -37,6 +37,7 @@ from agenttalk.store import (
     LEAD_LOOP_LEASE_ENV,
     LEAD_LOOP_REMINDER_AFTER_DEFAULT,
     OPENER_KINDS,
+    Message,
     Store,
     _owner_identity_gone,
     find_root,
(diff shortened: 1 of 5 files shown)


zoolok17:
Connector P2 disposition: correct finding, deliberately out of this PR's scope. It is exactly the Tier 3 leg dev-2's diagnosis identified and reviewer-3 approved deferring (supervisor.build_report is shared by the supervisor command and the wrapper loop, so threading the snapshot there needs its own caller audit). Now tracked as task #211 with the connector's sharpened trigger condition (active ephemeral reviewer) and the required scan-count-test fixture extension. This PR still eliminates the unconditional 24-scan cost; the residual is conditional on an active review being in flight.

